PR #40, opened 2019-05-16
Monero

hyc:
Tweaks to build within Monero source tree:

Add a CMakeLists.txt file for Monero's cmake setup
Rename some argon2 functions to avoid clash with libsodium


hyc:
After this is merged I'll be able to PR the other stuff to the Monero repo...

PR #41, opened 2019-05-18
Added POWER7+ intrinsincs and cpu detection to makefile

nioroso-x3:
Added altivec functions to support future jit compiler. Sadly IBM's AES implementation is not compatible with intels, see https://stackoverflow.com/questions/56027569/is-there-a-fast-way-to-make-ibms-vncipher-instruction-result-the-same-as-intels

### DIFF
--- a/makefile
+++ b/makefile
@@ -22,6 +22,14 @@ ifeq ($(PLATFORM),x86_64)
     CXXFLAGS += -maes
 endif
 
+ifeq ($(PLATFORM),ppc64)
+    CXXFLAGS += -mcpu=native
+endif
+
+ifeq ($(PLATFORM),ppc64le)
+    CXXFLAGS += -mcpu=native
+endif
+
 release: CXXFLAGS += -O3 -flto
 release: CCFLAGS += -O3 -flto
 release: LDFLAGS += -flto
--- a/src/intrin_portable.h
+++ b/src/intrin_portable.h
@@ -151,7 +151,225 @@ FORCE_INLINE void rx_set_rounding_mode(uint32_t mode) {
 	_mm_setcsr(rx_mxcsr_default | (mode << 13));
 }
 
+#elif defined(__PPC64__) && defined(__ALTIVEC__) && defined(__VSX__) //sadly only POWER7 and newer will be able to use SIMD acceleration. Earlier processors cant use doubles or 64 bit integers with SIMD
+#include <cstdint>
+#include <stdexcept>
+#include <cstdlib>
+#include<altivec.h>
+#undef vector
+#undef pixel
+#undef bool
+
+typedef __vector uint8_t __m128i;
+typedef __vector uint32_t __m128l;
+typedef __vector int      __m128li;
+typedef __vector uint64_t __m128ll;
+typedef __vector double __m128d;
+
+typedef __m128i rx_vec_i128;
+typedef __m128d rx_vec_f128;
+typedef union{
+	rx_vec_i128 i;
+  rx_vec_f128 d;
+  uint64_t u64[2];
+  double   d64[2];
+  uint32_t u32[4];
+	int i32[4];
+} vec_u;
+
+#define rx_aligned_alloc(a, b) malloc(a)
+#define rx_aligned_free(a) free(a)
+#define rx_prefetch_nta(x)
+
+
+/* Splat 64-bit long long to 2 64-bit long longs */
+FORCE_INLINE __m128i vec_splat2sd (int64_t scalar)
+{ return (__m128i) vec_splats (scalar); }
+
+FORCE_INLINE rx_vec_f128 rx_load_vec_f128(const double* pd) {
+#if defined(NATIVE_LITTLE_ENDIAN)
+	return (rx_vec_f128)vec_vsx_ld(0,pd);
 #else
+	vec_u t;
+	t.u64[0] = load64(pd + 0);
+	t.u64[1] = load64(pd + 1);
+	return (rx_vec_f128)t.d;
+#endif
+}
+
+FORCE_INLINE void rx_store_vec_f128(double* mem_addr, rx_vec_f128 a) {
+#if defined(NATIVE_LITTLE_ENDIAN)
+	vec_vsx_st(a,0,(rx_vec_f128*)mem_addr);
+#else
+	vec_u _a;
+	_a.d = a;
+	store64(mem_addr + 0, _a.u64[0]);
+	store64(mem_addr + 1, _a.u64[1]);
+#endif
+}
+
+FORCE_INLINE rx_vec_f128 rx_swap_vec_f128(rx_vec_f128 a) {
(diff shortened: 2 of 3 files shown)


tevador:
> Sadly IBM's AES implementation is not compatible with intels

It can still be used, you just have to pass a zero key and do the XOR separately at the end. The same thing has to be done on ARM.

For example (aarch64):

```c++
rx_vec_i128 rx_aesenc_vec_i128(rx_vec_i128 a, rx_vec_i128 key)
{
	const uint8x16_t zero = { 0 };
	return vaesmcq_u8(vaeseq_u8(a, zero)) ^ key;
}
```

BTW I can only merge your changes if you release your code under the 3-clause BSD license because RandomX will be relicensed.

nioroso-x3:
Thanks for the suggestion, managed to fix it using a zero roundkey and xoring with the real key at the end.
Also, I have no problems with the code being licensed under BSD.

tevador:
Could you post some benchmark comparison with and without this patch?

nioroso-x3:
Tests done on 8 threads of a single POWER8 core at 2.8ghz.

No altivec:
jribeiro@venetian:~/Development/RandomX$ taskset -c 8-15 ./bin/benchmark --verify --threads 8 --init 8 --nonces 160 --softAes
RandomX benchmark
 - light memory mode (256 MiB)
 - interpreted mode
 - software AES mode
 - small pages mode
Initializing ...
Memory initialized in 1.2905 s
Initializing 8 virtual machine(s) ...
Running benchmark (160 nonces) ...
Calculated result: 5b8edcc7a5f235d193397df02a53269cf060ca489805556f548fdce4e8766d6c
Performance: 264.217 ms per hash
jribeiro@venetian:~/Development/RandomX$ taskset -c 8-15 ./bin/benchmark --mine --threads 8 --init 8 --nonces 160 --softAes                                   
RandomX benchmark
 - full memory mode (2080 MiB)
 - interpreted mode
 - software AES mode
 - small pages mode
Initializing (8 threads) ...
Memory initialized in 509.995 s
Initializing 8 virtual machine(s) ...
Running benchmark (160 nonces) ...
Calculated result: 5b8edcc7a5f235d193397df02a53269cf060ca489805556f548fdce4e8766d6c
Performance: 47.4883 hashes per second
jribeiro@venetian:~/Development/RandomX$ 

With altivec:
jribeiro@venetian:/home/jribeiro/Development/RandomX-alti$ taskset -c 8-15 ./bin/benchmark --verify --threads 8 --init 8 --nonces 160
RandomX benchmark
 - light memory mode (256 MiB)
 - interpreted mode
 - hardware AES mode
 - small pages mode
Initializing ...
Memory initialized in 1.3005 s
Initializing 8 virtual machine(s) ...
Running benchmark (160 nonces) ...
Calculated result: 5b8edcc7a5f235d193397df02a53269cf060ca489805556f548fdce4e8766d6c
Performance: 236.58 ms per hash
jribeiro@venetian:/home/jribeiro/Development/RandomX-alti$ taskset -c 8-15 ./bin/benchmark --mine --threads 8 --init 8 --nonces 160                          
RandomX benchmark
 - full memory mode (2080 MiB)
 - interpreted mode
 - hardware AES mode
 - small pages mode
Initializing (8 threads) ...
Memory initialized in 462.852 s
Initializing 8 virtual machine(s) ...
Running benchmark (160 nonces) ...
Calculated result: 5b8edcc7a5f235d193397df02a53269cf060ca489805556f548fdce4e8766d6c
Performance: 54.9316 hashes per second

makefile
```diff
@@ -22,6 +22,18 @@ ifeq ($(PLATFORM),x86_64)
     CXXFLAGS += -maes
 endif
 
+ifeq ($(PLATFORM),ppc64)
+VSX=$(shell $(CC) -mcpu=native -dM -E - < /dev/null | grep VSX )
```
tevador: Is running this command necessary? What would happen if you always set `-mcpu=native -flax-vector-conversions` regardless of VSX support?

Also the GCC documentation says:
> **-flax-vector-conversions**
> Allow implicit conversions between vectors with differing numbers of elements and/or incompatible element types. This option should not be used for new code.

So it's best to remove the option and use explicit conversions.

src/intrin_portable.h
```diff
@@ -151,6 +151,211 @@ FORCE_INLINE void rx_set_rounding_mode(uint32_t mode) {
...
+}
+
+FORCE_INLINE rx_vec_i128 rx_aesdec_vec_i128(rx_vec_i128 v, rx_vec_i128 rkey) {
+	throw std::runtime_error(platformError);
+}
+#endif
+
+
+FORCE_INLINE int rx_vec_i128_x(rx_vec_i128 a) {
+	return ((int*)&a)[0];
```
tevador: It's better to do this using a `union` to prevent aliasing warnings from the compiler.

```c++
union vec_i32 {
    rx_vec_i128 vec;
    int i32[4];
};
```

src/intrin_portable.h
```diff
@@ -151,6 +151,211 @@ FORCE_INLINE void rx_set_rounding_mode(uint32_t mode) {
...
+	return vec_xor(_A,_B);
+}
+
+FORCE_INLINE rx_vec_i128 rx_load_vec_i128(rx_vec_i128 const*_P) {
+#if defined(NATIVE_LITTLE_ENDIAN)
+	return *_P;
+#else
+	uint32_t* ptr = (uint32_t*)_P;
+	uint32_t cu32[] = {0,0,0,0};
+  cu32[0] = load32(ptr + 0);
```
tevador: Broken indentation. Please use a single tab.

src/intrin_portable.h
```diff
@@ -151,6 +151,211 @@ FORCE_INLINE void rx_set_rounding_mode(uint32_t mode) {
...
+#include <cstdint>
+#include <stdexcept>
+#include <cstdlib>
+#include<altivec.h>
+#undef vector
+#undef pixel
+#undef bool
+
+typedef __vector uint8_t __m128i;
+typedef __vector uint16_t __m128h;
```
tevador: Unused typedef

makefile
nioroso-x3: the VSX define is set by the cpu type, -flax-vector-conversions shoudn't change it.
Ok, I'll change the code to use explicit conversions.

nioroso-x3:
Cleaned up with your suggestions.

makefile
tevador: I recommend removing this line: `VSX=$(shell $(CC) -mcpu=native -dM -E - < /dev/null | grep VSX )`.

So the script should be:
```makefile
ifeq ($(PLATFORM),ppc64)
    CXXFLAGS += -mcpu=native
endif
```

Additionally, please also modify the CMakeLists.txt file accordingly. Merge #56 to avoid conflicts.

nioroso-x3: Ok, merged #56 and fixed up the makefile.

PR #46, opened 2019-05-27
Use 4 AES rounds for program generation

tevador:
This improves the quality of the RNG without any measurable impact on performance.

Note: Scratchpad initialization still uses only 1 round for performance reasons.


src/aes_hash.cpp
```diff
@@ -143,3 +148,54 @@ void fillAes1Rx4(void *state, size_t outputSize, void *buffer) {
...
+	state1 = rx_load_vec_i128((rx_vec_i128*)state + 1);
+	state2 = rx_load_vec_i128((rx_vec_i128*)state + 2);
+	state3 = rx_load_vec_i128((rx_vec_i128*)state + 3);
+
+	while (outptr < outputEnd) {
+		state0 = aesdec<softAes>(state0, key0);
+		state1 = aesenc<softAes>(state1, key0);
+		state2 = aesdec<softAes>(state2, key0);
+		state3 = aesenc<softAes>(state3, key0);
+
```
SChernykh: state0 is encrypted with k0->k0->...->k0->k0->k1->k2->k3->k0->k1->... sequence
state1 is encrypted with k1->k1->...->k1->k0->k1->k2->k3->k0->k1->... sequence
state2 is encrypted with k2->k2->...->k2->k0->k1->k2->k3->k0->k1->... sequence
state3 is encrypted with k3->k3->...->k3->k0->k1->k2->k3->k0->k1->... sequence

It doesn't seem safe to me that k1 is used twice in "k1->k0->k1" sequence for example. I prefer to separate two uses of the same key when we start filling entropy. Maybe change key sequences so they don't all reset to k0 at the same iteration, i.e. do
```
	state0 = aesdec<softAes>(state0, key0);
	state1 = aesenc<softAes>(state1, key1);
	state2 = aesdec<softAes>(state2, key2);
	state3 = aesenc<softAes>(state3, key3);

	state0 = aesdec<softAes>(state0, key1);
	state1 = aesenc<softAes>(state1, key2);
	state2 = aesdec<softAes>(state2, key3);
	state3 = aesenc<softAes>(state3, key0);
```
and so on.

tevador: OK, we can rotate the keys, but I don't think there is any difference in security. The 1-round generator always reuses the same key anyways.

Or we could generate a completely new set of 4 round keys for this generator.

tevador: I changed AesGenerator4R to use a different set of round keys than AesGenerator1R.

PR #52, opened 2019-06-05
Add Dataset prefetch in interpreted VM

tevador:
* to formally match the specification
* a small increase in interpreted mining speed (~4%)


tevador:
As suggested in https://github.com/monero-project/monero/pull/5549#issuecomment-498881384

PR #58, opened 2019-06-08
Fix Windows detection

hyc:


tevador:
If I apply this patch, I get crashes when running the benchmark.

MINGW64 + gcc 7.4.0
```
user@machine MINGW64 ~/RandomX
$ gcc --version
gcc (GCC) 7.4.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
It seems this compiler doesn't define `__WIN32__`:
```
user@machine MINGW64 ~/RandomX
$ gcc -E -dM - < /dev/null | grep WIN
#define __WINT_MAX__ 0xffffffffU
#define __WINT_MIN__ 0U
#define __SIZEOF_WINT_T__ 4
#define __CYGWIN__ 1
#define __WINT_TYPE__ unsigned int
#define __WINT_WIDTH__ 32
```

If I apply this patch:
```
user@machine MINGW64 ~/RandomX
$ curl -L https://github.com/tevador/RandomX/pull/58.patch | git apply -v -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   140    0   140    0     0    469      0 --:--:-- --:--:-- --:--:--   477
100  1499    0  1499    0     0   2153      0 --:--:-- --:--:-- --:--:--  2153
Checking patch src/jit_compiler_x86_static.S...
Applied patch src/jit_compiler_x86_static.S cleanly.
```
Build, and run the benchmark:
```
user@machine MINGW64 ~/RandomX
$ bin/benchmark.exe --verify --jit
RandomX benchmark
 - light memory mode (256 MiB)
 - JIT compiled mode
 - hardware AES mode
 - small pages mode
Initializing ...
Memory initialized in 1.06841 s
Segmentation fault (core dumped)
```
Runs fine without this patch.

Also the `__CYGWIN__` macro is tested in [virtual_memory.cpp](https://github.com/tevador/RandomX/blob/master/src/virtual_memory.cpp#L33), so whatever fix is applied here should probably also be done there.

hyc:
virtual_memory.cpp tests both WIN32 and CYGWIN so it's correct.

tevador:
So we should check for both also in the assembly file. Something like:
```c
#if defined(__WIN32__) || defined(__CYGWIN__)
#define WINDOWS
#endif
```
And then change the conditions to `#if defined(WINDOWS)`.

PR #68, opened 2019-06-14
benchmark: add thread affinity mask option

jtgrassie:
Adds a thread affinity mask option for allowing to set miner threads to specific cpus/cores. Accepts a hex or decimal value representing a bitmask.

E.g. `--threads 4 --affinity 0x1D`, equates to our 4 threads `t1,t2,t3,t4` and a mask `11101`, which yields: ` (t4, cpu4), (t3, cpu3), (t2, cpu2), (t1, cpu0)`. So cpu is the bit position (right-to-left) and each `1` sequentially is the next thread.

Useful for benchmarking multi-processor/NUMA node systems.

### DIFF
--- a/makefile
+++ b/makefile
@@ -2,6 +2,7 @@
 #CC=gcc-8
 AR=gcc-ar
 PLATFORM=$(shell uname -m)
+OS=$(shell uname -s)
 CXXFLAGS=-std=c++11
 CCFLAGS=-std=c99
 ARFLAGS=rcs
@@ -20,6 +21,9 @@ endif
 ifeq ($(PLATFORM),x86_64)
     RXOBJS += $(addprefix $(OBJDIR)/,jit_compiler_x86_static.o jit_compiler_x86.o)
     CXXFLAGS += -maes
+endif
+ifeq ($(OS),Darwin)
+    AR=ar
 endif
 
 release: CXXFLAGS += -O3 -flto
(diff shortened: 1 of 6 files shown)


tevador:
Good work.

However, to take full advantage of NUMA, we would have to also support having multiple datasets. It may be possible to control dataset allocation with CPU affinity because operating systems tend to allocate memory in the node that first writes to the memory page (this would have to be confirmed).

jtgrassie:
Yes, that's my next task. I'm going to start looking into/testing hwloc for pinning caches and datasets to NUMA nodes and add to this PR.

jtgrassie:
> to take full advantage of NUMA, we would have to also support having multiple datasets

hwloc branch here https://github.com/jtgrassie/RandomX/tree/numa that uses multiple datasets and binds their memory to NUMA nodes. Then I'm using the affinity flag to bind threads to nodes. I'm not seeing a significant difference though, so need to investigate further before a PR of that hwloc branch.

hyc:
Fwiw, I ran multiple instances on separate NUMA nodes using numactl, don't recall seeing much impact then either.

tevador:
@jtgrassie FYI: I moved the affinity files into the 'tests' directory, since it's not directly related to the RandomX library. Also the `mask_to_string` function was giving some strange output starting with a null character, so I changed it to use `std::ostringstream`.

jtgrassie:
@tevador what was the issue with mask_to_string? I tested this pretty thoroughly and no issues. the nulls just present after the 1's and 0's to terminate the string. Also, would have been nice to preserve the files formatting (space vs tabs) as it now has a mixture.

tevador:
When built with Visual Studio, the output looked like this (for example `--affinity 5`):
```
 - thread affinity ( 101                                                            )
```
The whitespace chatacters inside the parentheses are actually null characters.

jtgrassie:
Very weird MS behavior of printing a string with nulls then! Can you fix the indentation on the file to keep consistently though please.

Issue #71, opened 2019-06-16
Can't Build in FreeNAS Jail


russoj88:
RE: https://www.reddit.com/r/MoneroMining/comments/c0ysjp/share_your_randomx_benchmarks/ercp7zq?utm_source=share&utm_medium=web2x

I'm trying to build this on a FreeNAS box, in a jail.
The jail's version: `FreeBSD 11.2-STABLE`
```shell
root@randomX:~ # gcc --version
gcc (FreeBSD Ports Collection) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```shell
root@randomX:~/RandomX # git checkout v1.0.3
HEAD is now at e4b2270 Merge pull request #47 from tevador/pr-doc
root@randomX:~/RandomX # make
make: "/root/RandomX/makefile" line 16: Missing dependency operator
make: "/root/RandomX/makefile" line 19: Need an operator
make: "/root/RandomX/makefile" line 20: Missing dependency operator
make: "/root/RandomX/makefile" line 23: Need an operator
make: "/root/RandomX/makefile" line 162: warning: duplicate script for target "obj/%.o" ignored
make: "/root/RandomX/makefile" line 159: warning: using previous script for "obj/%.o" defined here
make: "/root/RandomX/makefile" line 165: warning: duplicate script for target "obj/%.o" ignored
make: "/root/RandomX/makefile" line 159: warning: using previous script for "obj/%.o" defined here
make: Fatal errors encountered -- cannot continue
make: stopped in /root/RandomX
```

I get slightly farther with gmake:
```shell
root@randomX:~/RandomX # gmake --version
GNU Make 4.2.1
Built for amd64-portbld-freebsd11.2
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

root@randomX:~/RandomX # gmake
c++ -std=c++11 -maes -O3 -flto -c src/virtual_memory.cpp -o obj/virtual_memory.o
src/virtual_memory.cpp:119:83: error: use of undeclared identifier 'MAP_HUGETLB'
[...]

PR #74, opened 2019-06-19
Large page support for FreeBSD

tevador:


hyc:
I thought you decided this wasn't working? https://github.com/tevador/RandomX/issues/71#issuecomment-502865874

tevador:
It's needed because RandomX currently doesn't compile on FreeBSD. This patch has been tested to compile and run in #71.

hyc:
Sure it compiles and runs, but it doesn't seem to do anything useful wrt large pages. Maybe better to just  do
````
#ifdef __FreeBSD__
  return NULL;
#endif

**Comment by tevador** [comment]
I don't see a reason to disable large page allocation on FreeBSD based on one incomplete hashrate report. Do we know for sure it's not working?

BTW xmrig uses a similar way to allocate memory on FreeBSD: https://github.com/xmrig/xmrig/blob/master/src/Mem_unix.cpp#L56 

It uses the `MAP_PREFAULT_READ` flag, but I don't think it's strictly necessary based on [man](https://www.freebsd.org/cgi/man.cgi?mmap):

```
MAP_PREFAULT_READ	Immediately update the calling process's lowest-level
			virtual	address	translation structures,	such as	its
			page table, so that every memory resident page within
			the region is mapped for read access.  Ordinarily
			these structures are updated lazily.  The effect of
			this option is to eliminate any	soft faults that would
			otherwise occur	on the initial read accesses to	the
			region.	 Although this option does not preclude	prot
			from including PROT_WRITE, it does not eliminate soft
			faults on the initial write accesses to	the region.
```
The first access to each large page in RandomX is a write anyways.

PR #77, opened 2019-06-21
Decrease the frequency of FADD/FSUB in favor of FMUL

tevador:
* this better matches CPU capabilities since execution ports are usually split 1:1 between fadd and fmul


tevador:
* the frequency of FSWAP_R decreased from 8 to 4 (it's ASIC-friendly)

tevador:
I have also activated the IROL_R instruction at a 1:4 ratio with IROR_R (rotl is slower than rotr on ARM). The instruction should be active so that's covered by tests.

PR #87, opened 2019-06-25
Fix: possible overflow when using non-standard Dataset size

tevador:
Overflow is possible here if Dataset size is over 4 GiB.

https://github.com/tevador/RandomX/blob/2706a8b7533860900b33ac69899471e5127a4872/src/randomx.cpp#L123

Also added a `static_assert` to limit Dataset to a maximum of 16 GiB.

### DIFF
--- a/src/common.hpp
+++ b/src/common.hpp
@@ -43,6 +43,7 @@ namespace randomx {
 	static_assert((RANDOMX_DATASET_BASE_SIZE & (RANDOMX_DATASET_BASE_SIZE - 1)) == 0, "RANDOMX_DATASET_BASE_SIZE must be a power of 2.");
 	static_assert(RANDOMX_DATASET_BASE_SIZE <= 4294967296ULL, "RANDOMX_DATASET_BASE_SIZE must not exceed 4294967296.");
 	static_assert(RANDOMX_DATASET_EXTRA_SIZE % 64 == 0, "RANDOMX_DATASET_EXTRA_SIZE must be divisible by 64.");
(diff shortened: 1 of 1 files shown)


src/common.hpp
```diff
@@ -43,6 +43,7 @@ namespace randomx {
 	static_assert((RANDOMX_DATASET_BASE_SIZE & (RANDOMX_DATASET_BASE_SIZE - 1)) == 0, "RANDOMX_DATASET_BASE_SIZE must be a power of 2.");
 	static_assert(RANDOMX_DATASET_BASE_SIZE <= 4294967296ULL, "RANDOMX_DATASET_BASE_SIZE must not exceed 4294967296.");
 	static_assert(RANDOMX_DATASET_EXTRA_SIZE % 64 == 0, "RANDOMX_DATASET_EXTRA_SIZE must be divisible by 64.");
+	static_assert((uint64_t)RANDOMX_DATASET_BASE_SIZE + RANDOMX_DATASET_EXTRA_SIZE <= 17179869184, "Dataset size must not exceed 16 GiB.");
```
hyc: Is 16GB an arbitrary limit? Why do we need this limit?

tevador: The theoretical maximum is 256 GiB, because at that point dataset indexes will start to overflow.

But the maximum portion of the Dataset that is used by a program is 4 GiB, so it doesn't make much sense to have a very large Dataset. 16 GiB is an arbitrary limit to make sure RandomX is not used with broken parameters.

PR #90, opened 2019-06-27
Use cmake for building

tevador:
* CMakeLists.txt modified to support full standalone build
* added support for ARCH=native builds
* added PowerPC flags
* added ARMv8 flags
* check for x86 AES-NI at compile time


tevador:
I have also tested this with the Monero build.

zeno39:
Hi tevador, I have a question. I have a AMD epic 7351p 16core 32 thread 8 channels.
If i launch the script randomX i have juste 4000h/s and max 5000
I have 256 go in ddr4 i think it's not a problem, do you have any idea ? Probably m'y configuration is not good

tevador:
> Hi tevador, I have a question. I have a AMD epic 7351p 16core 32 thread 8 channels.
If i launch the script randomX i have juste 4000h/s and max 5000
I have 256 go in ddr4 i think it's not a problem, do you have any idea ? Probably m'y configuration is not good

Can you open a new issue? This discussion is for code review.

PR #92, opened 2019-06-28
Fix 'runtime_error' is not a member of 'std'

hyc:


tevador:
This PR should be against the pr-cmake3 branch and the include should be on line 589, otherwise it's not needed.

tevador:
This change is already included in #90.

PR #99, opened 2019-07-03
Do not attempt to allocate >= 4 GiB on 32-bit systems

tevador:
+ Cache size limited to 2 GiB

### DIFF
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -34,6 +34,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "vm_compiled_light.hpp"
 #include "blake2/blake2.h"
 #include <cassert>
+#include <limits>
 
 extern "C" {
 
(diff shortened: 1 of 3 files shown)


src/randomx.cpp
```diff
@@ -102,6 +103,12 @@ extern "C" {
 	}
 
 	randomx_dataset *randomx_alloc_dataset(randomx_flags flags) {
+
+		//fail on 32-bit systems if DatasetSize is >= 4 GiB
```
hyc: Shouldn't the failure threshold be much smaller? At 4GiB there is no room in the address space for any other code or data.

tevador: If the computer doesn't have enough virtual memory (which is likely) then the allocation will fail naturally.

This condition is in place to prevent an overflow which could make the allocation to falsely succeed. For example if `DatasetSize` is 4.5 GiB, it will overflow to 0.5 GiB on 32-bit systems and only 0.5 GiB will be allocated, leading to runtime crashes.

hyc: ok

PR #124, opened 2019-09-11
Optimized loading from scratchpad

SChernykh:
Prefetches scratchpad data as soon as it's possible to calculate data address for the next load.

~1.2% speedup on my Ryzen 7 3700X (from 8707 h/s up to 8814 h/s).


tevador:
This optimization almost exactly compensates the slowdown caused by #118

PR #127, opened 2019-09-23
Set scratchpad pointer to null by default

SChernykh:


tevador:
Good catch.

The problem is this delete:

https://github.com/tevador/RandomX/blob/67010ab5546abe8a2793992e4322ddfea021538e/src/randomx.cpp#L285-L288

PR #128, opened 2019-09-23
Workaround for a bug in binutils-2.32-1 on ARM

SChernykh:
ldr/madd instruction sequence makes compiled binary crash, so separate them.


hyc:
For future reference: was actually a hardware bug in revision 0 of ARM Cortex-A53, as noted here https://github.com/SChernykh/p2pool/pull/189

PR #130, opened 2019-09-25
Fix for win32 build with gcc

hyc:
Compiling with gcc on win32 is broken. This breaks gitian/depends builds of Monero.

### DIFF
--- a/src/instructions_portable.cpp
+++ b/src/instructions_portable.cpp
@@ -161,7 +161,7 @@ void rx_set_rounding_mode(uint32_t mode) {
 
 #ifdef RANDOMX_USE_X87
 
-#ifdef _M_IX86
+#if defined(_MSC_VER) && defined(_M_IX86)
 
 void rx_set_double_precision() {
 	_control87(_PC_53, _MCW_PC);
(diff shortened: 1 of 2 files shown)


tevador:
I'm not sure why GCC defines a MSVC-specific macro. https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019#microsoft-specific-predefined-macros

A simpler fix would be:
```
#if defined(_MSC_VER) && defined(_M_IX86)
```
because the whole purpose of the asm function is to fix a bug in the MSVC `sqrt` function. It shouldn't be required for GCC.

In any case, please run `randomx-tests` to verify that all rounding modes are working correctly.

hyc:
When building on Windows, all binaries are linked against the MSVC C runtime library, so if the bug is in the library as your comment indicates, then this is still needed even with gcc.

hyc:
64bit randomx-tests ran fine. I need to rebuild the 32bit test...

tevador:
> binaries are linked against the MSVC C runtime library

OK, then the assembly code is also needed. Let me know if the 32-bit tests pass, then I will merge it.

hyc:
No looks like you're right. The tests with my assembly patch failed. Just #if'ing it out works though. Then the tests proceed until test 84, compiler hash tests. I presume these are meant to fail since there is no 32bit JIT compiler.

tevador:
> tests proceed until test 84

Then it's OK.

There was a bug in the tests. It should skip the compiler hash tests if JIT is not supported. Fixed in 7c405a7

PR #131, opened 2019-09-28
CMake install target for randomx library

xiphon:
Needed for Monero GUI builds, we need the output library to link against it.


hyc:
Why does the GUI need the PoW algorithm?

xiphon:
> Why does the GUI need the PoW algorithm?

The GUI doesn't invoke any POW-related code directly. But ...

For example, to construct a tx GUI depends on `cryptonote_core`. And `cryptonote_core` depends on `cncrypto` that depends on `randomx`.

PR #133, opened 2019-10-03
Remember cache input data

hyc:
Make init_cache and set_cache no-ops if fed the same data as before

This is just a suggestion; it could make some code in monerod a tiny bit simpler if we
didn't have to track cache contents so closely. (E.g. https://github.com/monero-project/monero/pull/5948) Feel free to close if not wanted.

### DIFF
--- a/src/dataset.hpp
+++ b/src/dataset.hpp
@@ -50,7 +50,7 @@ struct randomx_cache {
 	randomx::DatasetInitFunc* datasetInit;
 	randomx::SuperscalarProgram programs[RANDOMX_CACHE_ACCESSES];
 	std::vector<uint64_t> reciprocalCache;
-	std::string cacheData;
+	std::string cacheKey;
 
 	bool isInitialized() {
 		return programs[0].getSize() != 0;
--- a/src/dataset.hpp
+++ b/src/dataset.hpp
@@ -50,6 +50,7 @@ struct randomx_cache {
 	randomx::DatasetInitFunc* datasetInit;
 	randomx::SuperscalarProgram programs[RANDOMX_CACHE_ACCESSES];
 	std::vector<uint64_t> reciprocalCache;
+	std::string cacheData;
 
 	bool isInitialized() {
 		return programs[0].getSize() != 0;
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -93,11 +93,11 @@ extern "C" {
 	void randomx_init_cache(randomx_cache *cache, const void *key, size_t keySize) {
 		assert(cache != nullptr);
 		assert(keySize == 0 || key != nullptr);
-		std::string data;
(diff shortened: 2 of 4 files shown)


src/dataset.hpp
```diff
@@ -50,6 +50,7 @@ struct randomx_cache {
 	randomx::DatasetInitFunc* datasetInit;
 	randomx::SuperscalarProgram programs[RANDOMX_CACHE_ACCESSES];
 	std::vector<uint64_t> reciprocalCache;
+	std::string cacheData;
```
tevador: Should be called `cacheKey` or there should be a comment clarifying that it's the key that was used for initialization (to avoid confusion with actual cache data).

src/randomx.cpp
```diff
@@ -93,7 +93,12 @@ extern "C" {
 	void randomx_init_cache(randomx_cache *cache, const void *key, size_t keySize) {
 		assert(cache != nullptr);
 		assert(keySize == 0 || key != nullptr);
-		cache->initialize(cache, key, keySize);
+		std::string data;
+		data.assign((const char *)key, keySize);
+		if (cache->cacheData != data) {
+			cache->initialize(cache, key, keySize);
+			cache->cacheData = data;
+		}
```
tevador: This will leave the cache uninitialized if the key is empty (which is a valid key value). You should add `|| !cache->isInitialized()`.

src/randomx.cpp
```diff
@@ -293,7 +298,10 @@ extern "C" {
 	void randomx_vm_set_cache(randomx_vm *machine, randomx_cache* cache) {
 		assert(machine != nullptr);
 		assert(cache != nullptr && cache->isInitialized());
-		machine->setCache(cache);
+		if (machine->cacheData != cache->cacheData) {
+			machine->setCache(cache);
+			machine->cacheData = cache->cacheData;
+		}
```
tevador: You need to set `machine->cacheData` also on line 283 when the VM is created.

src/virtual_machine.hpp
```diff
@@ -65,6 +65,8 @@ class randomx_vm {
 		randomx_dataset* datasetPtr;
 	};
 	uint64_t datasetOffset;
+public:
+	std::string cacheData;
```
tevador: I suggest to rename to `cacheKey`.

src/dataset.hpp
hyc: renamed

src/randomx.cpp
hyc: added isInit test

src/randomx.cpp
hyc: set

src/virtual_machine.hpp
hyc: renamed

PR #136, opened 2019-10-09
Automatic detection of CPU capabilities

tevador:
Can autodetect AES, SSSE3 and AVX2 and sets the JIT flag if supported.

Usage:
```c
randomx_flags flags = randomx_get_flags();
randomx_cache *myCache = randomx_alloc_cache(flags);
randomx_init_cache(myCache, &myKey, sizeof myKey);
randomx_vm *myMachine = randomx_create_vm(flags, myCache, NULL);
``` 

AES detection on ARM64 needs testing.

### DIFF
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -48,7 +48,7 @@ extern "C" {
 		if (randomx_argon2_impl_avx2() != nullptr && cpu.hasAvx2()) {
 			flags |= RANDOMX_FLAG_ARGON2_AVX2;
 		}
-		else if (randomx_argon2_impl_ssse3() != nullptr && cpu.hasSsse3()) {
+		if (randomx_argon2_impl_ssse3() != nullptr && cpu.hasSsse3()) {
 			flags |= RANDOMX_FLAG_ARGON2_SSSE3;
 		}
 		return flags;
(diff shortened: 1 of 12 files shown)


tevador:
BTW also includes `--auto` flag in the benchmark, so:
```
./randomx-benchmark --mine --jit --largePages --init 8 --avx2
```
can be simplified to
```
./randomx-benchmark --mine --auto --largePages
```

tevador:
@SChernykh please review

src/randomx.cpp
```diff
@@ -33,18 +33,38 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+			flags |= RANDOMX_FLAG_ARGON2_SSSE3;
+		}
+		return flags;
+	}
+
 	randomx_cache *randomx_alloc_cache(randomx_flags flags) {
 		randomx_cache *cache = nullptr;
+		auto impl = randomx::selectArgonImpl(flags);
+		if (impl == nullptr) {
+			return cache;
```
SChernykh: It's probably a bad idea to return uninitialized cache here even if impl can't be nullptr with current code. It should output some error and abort, or throw an exception.

src/randomx.cpp
```diff
@@ -33,18 +33,38 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+#include "cpu.hpp"
 #include <cassert>
 #include <limits>
 
 extern "C" {
 
+	randomx_flags randomx_get_flags() {
+		randomx_flags flags = RANDOMX_HAVE_COMPILER ? RANDOMX_FLAG_JIT : RANDOMX_FLAG_DEFAULT;
+		randomx::Cpu cpu;
+		if (HAVE_AES && cpu.hasAes()) {
```
SChernykh: It's better to wrap it into `#if HAVE_AES == 1 ... #endif`, no need to make it compile-time when preprocessor can handle it.

src/randomx.cpp
tevador: 1. `impl` can be null if the caller selects an unsupported implementation (e.g. AVX2 on ARM).
1. It's not returning an uninitialized cache but simply a NULL pointer. It's up to the client to check for NULLs.
1. This is a C-language API, so we can't throw exceptions unless we handle them internally.

SChernykh: Then it's ok, I somehow thought cache was already allocated at this point.

PR #139, opened 2019-10-17
Fixed mmap on OpenBSD

ston1th:
OpenBSD's mmap does not support `MAP_HUGETLB` and `MAP_POPULATE`.


tevador:
Is there no way to allocate huge pages on OpenBSD?

ston1th:
I dont think so.
The manpage does not mention both of them: https://man.openbsd.org/mmap

tevador:
Perhaps the best solution would be this:

```c++
#elif defined(__OpenBSD__)
      mem = nullptr; //hugepages not supported
```
Like this, the allocator would not lie about having successfully allocated large pages. In monerod, it would automatically fall back to the C++ allocator (and show a warning).

ston1th:
Sounds good. I will test this with monero.

ston1th:
Compilation and tests of monero are successful.

PR #140, opened 2019-10-17
Update CMakeLists.txt

Alkenso:
Changed file type of src/jit_compiler_a64_static.S to 'ASM'

At least required by Xcode


tevador:
I can't recall the exact reason, but we set those files to be treated as C code on purpose.
``` 
# cheat because cmake and ccache hate each other
```

Alkenso:
@tevador , updated exclusively for Xcode

PR #142, opened 2019-10-19
Fixed tests on OpenBSD

ston1th:
Since OpenBSD 6.0 W^X is enforced.
Added `RANDOMX_FLAG_SECURE` in tests and benchmarks.

Excluded `cpu_set_t` since it is not defined on OpenBSD.

### DIFF
--- a/src/tests/affinity.cpp
+++ b/src/tests/affinity.cpp
@@ -65,7 +65,7 @@ set_thread_affinity(std::thread::native_handle_type thread,
             (thread_policy_t)&policy, 1);
 #elif defined(_WIN32) || defined(__CYGWIN__)
     rc = SetThreadAffinityMask(reinterpret_cast<HANDLE>(thread), 1ULL << cpuid) == 0 ? -2 : 0;
-#else
+#elif !defined(__OpenBSD__)
     cpu_set_t cs;
     CPU_ZERO(&cs);
     CPU_SET(cpuid, &cs);
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -42,6 +42,11 @@ extern "C" {
 	randomx_flags randomx_get_flags() {
 		randomx_flags flags = RANDOMX_HAVE_COMPILER ? RANDOMX_FLAG_JIT : RANDOMX_FLAG_DEFAULT;
 		randomx::Cpu cpu;
+#ifdef __OpenBSD__
+		if (flags == RANDOMX_FLAG_JIT) {
+			flags |= RANDOMX_FLAG_SECURE;
+		}
+#endif
 		if (HAVE_AES && cpu.hasAes()) {
 			flags |= RANDOMX_FLAG_HARD_AES;
 		}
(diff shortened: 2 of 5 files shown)


tevador:
Do all tests pass on OpenBSD?

ston1th:
Yes they do.
Except for the AVX2 test, which is skipped because I do not have the required hardware.

tevador:
I'm thinking to add `RANDOMX_FLAG_SECURE` to `randomx_get_flags` on OpenBSD, since the purpose of this function is to return flags that will work 100% on the target system. I think if you try mining in monerod it will fail even with this patch.

https://github.com/tevador/RandomX/blob/4296c35d79e3ff3ba03977b0be16c6a24564096e/src/randomx.cpp#L42-L55

ston1th:
Good idea. I updated the patch to account for this.

It doesn't seem that monero code calls `randomx_get_flags()` anywhere, but I have an extra patch for monerod which enforces `RANDOMX_FLAG_SECURE` on OpenBSD when in JIT mode (mining or not): https://github.com/monero-project/monero/pull/6003

Despite the point I don't think OpenBSD is a system to mine efficiently on :)

tevador:
@ston1th Please also update the comment here

https://github.com/tevador/RandomX/blob/e0484dfb99adf98b2b2a17c01b0cd883e8d4e1f3/src/randomx.h#L84

It should say that it sets the flag on OpenBSD because it's enforced by the OS.

src/tests/affinity.cpp
```diff
@@ -65,7 +65,7 @@ set_thread_affinity(std::thread::native_handle_type thread,
             (thread_policy_t)&policy, 1);
 #elif defined(_WIN32) || defined(__CYGWIN__)
     rc = SetThreadAffinityMask(reinterpret_cast<HANDLE>(thread), 1ULL << cpuid) == 0 ? -2 : 0;
-#else
+#elif !defined(__OpenBSD__)
```
jtgrassie: What's wrong with this on OpenBSD?

jtgrassie: Nvm, I missed your comment.

SChernykh: This should actually be rewritten for OpenBSD: https://man.openbsd.org/NetBSD-7.0.1/affinity.3
Struct name is `cpuset_t` there.

SChernykh:
@ston1th 
`Excluded cpu_set_t since it is not defined on OpenBSD.`
Just add `#define cpu_set_t cpuset_t` for OpenBSD. It should work fine, just the name is different on OpenBSD.

src/tests/affinity.cpp
ston1th: I may be mistaken but you are referring to a man page of NetBSD.
I can't find either one of these in the OpenBSD sources: `cpu_set_t`, `cpuset_t`, `CPU_ZERO`, `CPU_SET`.

I think there is no thread affinity for OpenBSD: https://man.openbsd.org/pthreads.3

SChernykh: Yes, you're right and it seems to be a design decision for OpenBSD.

PR #144, opened 2019-10-24
Fixes for cmake build with visual studio

VladimirVR:
* Added project name to CMake.
* Added MASM file for Visual Studio builds instead of default asm.
* Replaced static_assert to warning message for Visual Studio Debug build.


src/dataset.hpp
```diff
@@ -62,7 +62,13 @@ struct randomx_cache {
 //A pointer to a standard-layout struct object points to its initial member
 static_assert(std::is_standard_layout<randomx_dataset>(), "randomx_dataset must be a standard-layout struct");
 //the following assert fails when compiling Debug in Visual Studio (JIT mode will crash in Debug)
+#if _MSC_VER && !__INTEL_COMPILER && _DEBUG
+#define _STR(x) #x
+#define STR(x) _STR(x)
+#pragma message ( __FILE__ "(" STR(__LINE__) ") warning: check std::is_standard_layout<randomx_cache>() is disabled for Debug configuration" )
```
SChernykh: This assert must be there because ASM code assumes that pointer to cache memory is the first member of randomx_cache struct, this is exactly what std::is_standard_layout checks.

Dev-vlad:
> JIT compiler doesn't work in Debug builds with Visual Studio, removing this static_assert won't make it work anyway.
> 
> P.S. This assert must be there because ASM code assumes that pointer to cache memory is the first member of randomx_cache struct, this is exactly what std::is_standard_layout checks.

The main intention of this change not to fix the runtime crash, this is obviously not so easy. It allows to pass build stage for projects that use library as a third party dependency.
I can extend warning message with info that this this disabled check causes crash.

tevador:
I'm OK with removing the assert in Debug configuration with a warning.

src/dataset.hpp
```diff
@@ -62,7 +62,13 @@ struct randomx_cache {
 //A pointer to a standard-layout struct object points to its initial member
 static_assert(std::is_standard_layout<randomx_dataset>(), "randomx_dataset must be a standard-layout struct");
 //the following assert fails when compiling Debug in Visual Studio (JIT mode will crash in Debug)
+#if _MSC_VER && !__INTEL_COMPILER && _DEBUG
```
tevador: This should be `#if defined(_MSC_VER) && !defined(__INTEL_COMPILER) && defined(_DEBUG)`

CMakeLists.txt
```diff
@@ -96,11 +98,20 @@ endfunction()
...
+    list(APPEND randomx_sources src/jit_compiler_x86_static.asm)
+
+    set_property(SOURCE src/jit_compiler_x86_static.asm PROPERTY LANGUAGE ASM_MASM)
+  else()
+    list(APPEND randomx_sources src/jit_compiler_x86_static.S)
+
+    # cheat because cmake and ccache hate each other
+    set_property(SOURCE src/jit_compiler_x86_static.S PROPERTY LANGUAGE C)
+    set_property(SOURCE src/jit_compiler_x86_static.S PROPERTY XCODE_EXPLICIT_FILE_TYPE sourcecode.asm)
+  endif()
```
tevador: You need to also explicitly enable AVX2 instruction set for src/argon2_avx2.c, otherwise AVX2 will not work.

src/dataset.hpp
```diff
@@ -62,7 +62,13 @@ struct randomx_cache {
...
+#define _STR(x) #x
```
tevador: Identifiers starting with an underscore are reserved for the compiler. Please rename.

CMakeLists.txt
Dev-vlad: Proper check for availability of AVX2 instructions for MSVC is not so trivial, because requires runtime check (check this gist https://gist.github.com/UnaNancyOwen/263c243ae1e05a2f9d0e).
I think this is out of scope for this change and should be done in a separate change.

src/dataset.hpp
Dev-vlad: Done

src/dataset.hpp
Dev-vlad: Done

CMakeLists.txt
tevador: RandomX already does a [runtime check](https://github.com/tevador/RandomX/blob/master/src/randomx.cpp#L53). It's enough to unconditionally compile src/argon2_avx2.c with AVX2 instructions turned on, just like [we do with GCC](https://github.com/tevador/RandomX/blob/master/CMakeLists.txt#L116).

Be careful to not enable AVX2 for the whole build. It's safe only for this one file (src/argon2_avx2.c).

Dev-vlad: Done

tevador:
There is one more issue with compiling in Visual Studio: the configuration must be converted to MASM format to be accessible in assembly.

At the moment, it's done using a powershell build script: https://github.com/tevador/RandomX/blob/master/vcxproj/randomx.vcxproj#L128

Not sure how to add this to the makefile, though.

Dev-vlad:
> There is one more issue with compiling in Visual Studio: the configuration must be converted to MASM format to be accessible in assembly.
> 
> At the moment, it's done using a powershell build script: https://github.com/tevador/RandomX/blob/master/vcxproj/randomx.vcxproj#L128
> 
> Not sure how to add this to the makefile, though.

I know, will add it to CMake.

VladimirVR:
@tevador Done

src/dataset.hpp
VladimirVR: Done (see comments below)

Alkenso:
Also waiting for this PR. Maybe some work/help is needed to complete it?

giaki3003:
PR seems ok, merge maybe? @tevador

tevador:
I will try to find time tomorrow to test it and merge.

In the meantime, you can apply the patch manually
```
curl -L https://github.com/tevador/RandomX/pull/144.patch | git apply -v -
```

giaki3003:
> I will try to find time tomorrow to test it and merge.
> 
> In the meantime, you can apply the patch manually
> 
> ```
> curl -L https://github.com/tevador/RandomX/pull/144.patch | git apply -v -
> ```

Oh yes, i've been doing that already. No worries

PR #160, opened 2019-11-19
Improve install target (fixed)

tevador:
This fixes #150, which was reverted.


ehoffman2:
I can confirm this is working in my setup.
It also fix issue I did not mentioned last time about warning of missing PUBLIC_HEADER destination.

Eric

PR #166, opened 2019-12-01
Combined hash and fill AES loop

SChernykh:
Adds more parallelizm into AES loop so modern CPUs can take advantage of it. Also, scratchpad data moves between L1 and L3 caches only one time which saves time and energy per hash.

### DIFF
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -364,20 +364,20 @@ extern "C" {
 	}
 
 	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t *tempHash, const void* input, size_t inputSize) {
-		blake2b(tempHash, sizeof(tempHash), input, inputSize, nullptr, 0);
+		blake2b(tempHash, sizeof(uint64_t) * 8, input, inputSize, nullptr, 0);
 		machine->initScratchpad(tempHash);
 	}
 
 	void randomx_calculate_hash_next(randomx_vm* machine, uint64_t *tempHash, const void* nextInput, size_t nextInputSize, void* output) {
 		machine->resetRoundingMode();
 		for (uint32_t chain = 0; chain < RANDOMX_PROGRAM_COUNT - 1; ++chain) {
-			machine->run(&tempHash);
-			blake2b(tempHash, sizeof(tempHash), machine->getRegisterFile(), sizeof(randomx::RegisterFile), nullptr, 0);
+			machine->run(tempHash);
+			blake2b(tempHash, sizeof(uint64_t) * 8, machine->getRegisterFile(), sizeof(randomx::RegisterFile), nullptr, 0);
 		}
-		machine->run(&tempHash);
+		machine->run(tempHash);
 
 		// Finish current hash and fill the scratchpad for the next hash at the same time
-		blake2b(tempHash, sizeof(tempHash), nextInput, nextInputSize, nullptr, 0);
+		blake2b(tempHash, sizeof(uint64_t) * 8, nextInput, nextInputSize, nullptr, 0);
 		machine->hashAndFill(output, RANDOMX_HASH_SIZE, tempHash);
 	}
 }
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -363,12 +363,12 @@ extern "C" {
 		machine->getFinalResult(output, RANDOMX_HASH_SIZE);
 	}
 
-	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t (&tempHash)[8], const void* input, size_t inputSize) {
+	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t *tempHash, const void* input, size_t inputSize) {
 		blake2b(tempHash, sizeof(tempHash), input, inputSize, nullptr, 0);
 		machine->initScratchpad(tempHash);
 	}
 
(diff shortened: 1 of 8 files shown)


src/randomx.h
```diff
@@ -238,6 +239,21 @@ RANDOMX_EXPORT void randomx_destroy_vm(randomx_vm *machine);
...
+ * @param tempHash an array of 8 64-bit values used to store intermediate data between calls to randomx_calculate_hash_first and randomx_calculate_hash_next.
+ * @param input is a pointer to memory to be hashed. Must not be NULL.
+ * @param inputSize is the number of bytes to be hashed.
+ * @param nextInput is a pointer to memory to be hashed for the next hash. Must not be NULL.
+ * @param nextInputSize is the number of bytes to be hashed for the next hash.
+ * @param output is a pointer to memory where the hash will be stored. Must not
+ *        be NULL and at least RANDOMX_HASH_SIZE bytes must be available for writing.
+*/
+RANDOMX_EXPORT void randomx_calculate_hash_first(randomx_vm* machine, uint64_t (&tempHash)[8], const void* input, size_t inputSize);
+RANDOMX_EXPORT void randomx_calculate_hash_next(randomx_vm* machine, uint64_t (&tempHash)[8], const void* nextInput, size_t nextInputSize, void* output);
```
tevador: `tempHash` must be a pointer. C doesn't have references (the API must be usable in pure C code).

SChernykh: Fixed

src/randomx.cpp
```diff
@@ -363,12 +363,12 @@ extern "C" {
 		machine->getFinalResult(output, RANDOMX_HASH_SIZE);
 	}
 
-	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t (&tempHash)[8], const void* input, size_t inputSize) {
+	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t *tempHash, const void* input, size_t inputSize) {
```
hyc: For typechecking purposes you could still typedef this param to `uint64_t foo[8]` and use the type in the parameter lists.

src/randomx.cpp
```diff
@@ -363,4 +363,21 @@ extern "C" {
...
 
+	void randomx_calculate_hash_first(randomx_vm* machine, uint64_t (&tempHash)[8], const void* input, size_t inputSize) {
+		blake2b(tempHash, sizeof(tempHash), input, inputSize, nullptr, 0);
+		machine->initScratchpad(tempHash);
+	}
+
+	void randomx_calculate_hash_next(randomx_vm* machine, uint64_t (&tempHash)[8], const void* nextInput, size_t nextInputSize, void* output) {
+		machine->resetRoundingMode();
+		for (uint32_t chain = 0; chain < RANDOMX_PROGRAM_COUNT - 1; ++chain) {
+			machine->run(&tempHash);
```
tevador: This causes a segfault after you changed `tempHash` to a pointer.

src/randomx.cpp
```diff
@@ -363,4 +363,21 @@ extern "C" {
...
+			blake2b(tempHash, sizeof(tempHash), machine->getRegisterFile(), sizeof(randomx::RegisterFile), nullptr, 0);
+		}
+		machine->run(&tempHash);
+
+		// Finish current hash and fill the scratchpad for the next hash at the same time
+		blake2b(tempHash, sizeof(tempHash), nextInput, nextInputSize, nullptr, 0);
```
tevador: `sizeof(tempHash)` must be fixed here, otherwise the hash is incorrect.

src/randomx.cpp
SChernykh: Fixed

src/randomx.cpp
SChernykh: Fixed

src/randomx.cpp
tevador: Yeah, `uint64_t tempHash[8]` could be more readable in the API and it decays into a pointer anyways.

hyc: And that would have prevented that sizeof() bug too.

maoxs2: Using the "first" and "next" hash function in randomx.cpp, the miner (mine which based on this repo, not xmirg) receives "Low difficulty share" from pool.

SChernykh: You have to be careful using these new functions, check how benchmark.cpp handles them and where it updates the nonce.

maoxs2: My miner works fine on this workflow 
```
	uint64_t tempHash[8];
	while (nonce < noncesCount) {
		nonce = atomicNonce.fetch_add(1);
		store32(noncePtr, nonce);
	        randomx_calculate_hash_first(vm, tempHash, blockTemplate, sizeof(blockTemplate));
		randomx_calculate_hash_next(vm, tempHash, blockTemplate, sizeof(blockTemplate), &hash);
		result.xorWith(hash);
	}
```

But when the `randomx_calculate_hash_first` func is out of while loop, the miner gets the wrong result.
```
	uint64_t tempHash[8];

	store32(noncePtr, nonce);
	randomx_calculate_hash_first(vm, tempHash, blockTemplate, sizeof(blockTemplate));

	while (nonce < noncesCount) {
		nonce = atomicNonce.fetch_add(1);
		store32(noncePtr, nonce);
		randomx_calculate_hash_next(vm, tempHash, blockTemplate, sizeof(blockTemplate), &hash);
		result.xorWith(hash);
	}
```

PR #171, opened 2019-12-01
Disable thread affinity on FreeBSD and Android

tevador:


hyc:
Probably FreeBSD just needs `#include <pthread_np.h>` to have it supported. According to https://www.unix.com/man-page/freebsd/3/pthread_affinity_np/

PR #183, opened 2020-05-28
virtual_memory: add MAP_JIT on macOS

selsta:
This change makes it possible to enable hardened runtime with `com.apple.security.cs.allow-jit` key instead of `com.apple.security.cs.allow-unsigned-executable-memory`.

A bit ugly with nested `#ifdef`. Apparently there is a crashing issue on macOS 10.13 or lower but I was not able to reproduce it on macOS 10.12: https://github.com/herumi/xbyak/pull/84#issuecomment-562442380

Also does this make issues on iOS or does it use interpreter mode anyway?

### DIFF
--- a/src/virtual_memory.cpp
+++ b/src/virtual_memory.cpp
@@ -99,7 +99,17 @@ void* allocMemoryPages(std::size_t bytes) {
 	#else
 		#define RESERVED_FLAGS 0
 	#endif
-	mem = mmap(nullptr, bytes, PAGE_READWRITE | RESERVED_FLAGS, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+	#ifdef __APPLE__
+		#include <TargetConditionals.h>
+		#ifdef TARGET_OS_OSX
+			#define MEXTRA MAP_JIT
+		#else
+			#define MEXTRA 0
+		#endif
+	#else
+		#define MEXTRA 0
+	#endif
+	mem = mmap(nullptr, bytes, PAGE_READWRITE | RESERVED_FLAGS, MAP_ANONYMOUS | MAP_PRIVATE | MEXTRA, -1, 0);
 	if (mem == MAP_FAILED)
 		throw std::runtime_error("allocMemoryPages - mmap failed");
 #endif


src/virtual_memory.cpp
```diff
@@ -93,8 +93,12 @@ void* allocMemoryPages(std::size_t bytes) {
 	mem = VirtualAlloc(nullptr, bytes, MEM_COMMIT, PAGE_READWRITE);
 	if (mem == nullptr)
 		throw std::runtime_error(getErrorMessage("allocMemoryPages - VirtualAlloc"));
+#else
+#ifdef __APPLE__
+	mem = mmap(nullptr, bytes, PAGE_READWRITE, MAP_ANONYMOUS | MAP_PRIVATE | MAP_JIT, -1, 0);
```
hyc, line 98: It would be a slight bit cleaner to do something like
```
#ifdef __APPLE__
#define MAPFLAGS MAP_ANONYMOUS|MAP_PRIVATE|MAP_JIT
#else 
#define MAPFLAGS MAP_ANONYMOUS|MAP_PRIVATE
#endif 
     mem = mmap(nullptr, bytes, PAGE_READWRITE, MAPFLAGS, -1, 0);
```

hyc: Another alternative
```
#ifdef __APPLE__
#define MEXTRA MAP_JIT
#else 
#define MEXTRA 0
#endif 
    mem = mmap(nullptr, bytes, PAGE_READWRITE, MAP_ANONYMOUS | MAP_PRIVATE | MEXTRA, -1, 0);

```

tevador:
> Also does this make issues on iOS or does it use interpreter mode anyway?

AFAIK you need a jailbroken device to run JIT on iOS. But please check that this change doesn't break the iOS build since the `__APPLE__` macro is for both macOS and iOS.

selsta:
@tevador Is there a documentation on how to do iOS builds?

tevador:
@selsta Can you please rebase? There are some conflicts now after merging the NetBSD patch.

selsta:
@tevador Updated, open for code style improvements, ifdefs are quite nested now.

src/virtual_memory.cpp
```diff
@@ -99,7 +99,17 @@ void* allocMemoryPages(std::size_t bytes) {
 	#else
 		#define RESERVED_FLAGS 0
 	#endif
-	mem = mmap(nullptr, bytes, PAGE_READWRITE | RESERVED_FLAGS, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+	#ifdef __APPLE__
+		#include <TargetConditionals.h>
```
hyc, line 103: Usually the '#' must be the first character on the line. If you want to indent, use something like "#    ifdef xxxx" instead

selsta: Updated.

tevador: @hyc That has not been the case since ANSI C AFAIK.

See for example http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf page 146 and onward.

tevador: @selsta I prefer if you restore the original indentation for readability.

hyc: This looks worse with no indentation at all. Might as well go back to your original indentation.

selsta: done

tevador (approved):

PR #188, opened 2020-06-27
replace hardcoded literal with its appropriate symbol

cryptonote-social:
I believe his hardcoded "8" should actually be RegistersCount, just in case anyone tries to create a variant with more/fewer registers.


hyc:
Your first commit looks correct. The second commit has nothing to do with the topic and should be in a separate PR.

cryptonote-social:
oops, yeah didn't mean to include that one. fixing...

cryptonote-social:
second commit reverted

hyc:
you should have just used git rebase to delete the unintended commit and then force-pushed your branch.

cryptonote-social:
done

PR #197, opened 2020-11-29
Invalidate instruction cache on apple silicon

tevador:


hyc:
A bit puzzled here too. @selsta says Intel MacOS doesn't require --secure, so not sure why ARM version requires it.

hyc:
How about calling sys_dcache_flush before sys_icache_invalidate ?

hyc:
Or using        sys_cache_control(kCacheFunctionPrepareForExecution, code, size);
(taken from https://opensource.apple.com/source/JavaScriptCore/JavaScriptCore-7603.3.8/assembler/ARM64Assembler.h.auto.html )

tevador:
[This developer](https://developer.apple.com/documentation/apple_silicon/porting_just-in-time_compilers_to_apple_silicon) guide says `sys_icache_invalidate` should be called. It also says that the Apple silicon devices enforce W^X, which explains why `--secure` is required. They mention a new function `pthread_jit_write_protect_np`, which is faster than `mprotect` when switching from RW to RX access, so it should be possible to improve the hashrate a bit if it's implemented in RandomX.

hyc:
Ok, so this is a basic difference between Intel and ARM Macs, got it.

hyc:
So this ought to work
[dif.txt](https://github.com/tevador/RandomX/files/5612867/dif.txt)

PR #199, opened 2020-11-30
Faster W^X policy for apple silicon macs

tevador:
Note: this patch is untested as I don't own an apple device.


hyc:
Running this PR gets a Bus error. Fails with or without --secure.

PR #202, opened 2021-02-01
Fix illegal instruction crash on some ARM systems

SChernykh:
See https://github.com/xmrig/xmrig/issues/1916

### DIFF
--- a/src/jit_compiler_a64.cpp
+++ b/src/jit_compiler_a64.cpp
@@ -95,6 +95,10 @@ JitCompilerA64::JitCompilerA64()
 {
 	memset(reg_changed_offset, 0, sizeof(reg_changed_offset));
 	memcpy(code, (void*) randomx_program_aarch64, CodeSize);
+
+#ifdef __GNUC__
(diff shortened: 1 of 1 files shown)


SChernykh:
@tevador Can you merge this?

src/jit_compiler_a64.cpp
```diff
@@ -95,6 +95,10 @@ JitCompilerA64::JitCompilerA64()
 {
 	memset(reg_changed_offset, 0, sizeof(reg_changed_offset));
 	memcpy(code, (void*) randomx_program_aarch64, CodeSize);
+
+#ifdef __GNUC__
```
hyc, line 99: Is this really gcc-specific? The Android SDK uses clang, will this work there, or is some other fix needed for an Android build?

SChernykh: It's a GCC-specific builtin, but clang supports it and even defines `__GNUC__`. Also, the same `#ifdef` is used in other places here.

Edan3blov:
Hello,

This code was added and compiled but still mining crashs on smarphones.

If JIT is disabled mining works but without JIT there is no serius mining.

PR #211, opened 2021-05-20
Optimized dataset read

SChernykh:
There was a false dependency on readReg2 and readReg3 (caused by `xor rbp, rax` instruction) when reading dataset item (see design.md - 4.6.2 Loop execution, steps 5 and 7). This change uses `ma` register to read dataset item before the whole `rbp` (`ma` and `mx`) is changed, so superscalar and out-of-order CPU can start executing it earlier.

Results: https://i.imgur.com/Bpeq9mx.png

~1% speedup on modern Intel/AMD CPUs.

### DIFF
--- a/src/asm/program_prologue_linux.inc
+++ b/src/asm/program_prologue_linux.inc
@@ -15,6 +15,7 @@
 	mov rsi, rdx                ;# uint8_t* scratchpad
 
 	mov rax, rbp
+	ror rbp, 32
 
 	;# zero integer registers
 	xor r8, r8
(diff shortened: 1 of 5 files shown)


tevador:
I can confirm a 0.6% speedup on Ryzen 3700X.

tevador:
However, if you run `randomx-tests`, test 84 fails with an invalid hash.

SChernykh:
This is weird. I got a correct hash after 100k iterations, see screenshot. What's in test 84?

SChernykh:
I've just double checked that randomx-benchmark.exe gives correct hashes. I'll look into it.

tevador:
Everything works now.

PR #222, opened 2021-11-04
Fix absolute address references

hyc:
Use local symbols instead of global symbols, to prevent problems
with dynamic linker.

Was causing reproducible crashes in monero core_tests.
The .S patches have been tested on ARM and x86-64 Linux. The corresponding .asm patches are untested and need review, perhaps they're not needed.

The crashes were only in debug builds; they use dynamically linked libraries. This bug only hits when librandomx is a shared library.

### DIFF
--- a/src/jit_compiler_x86_static.asm
+++ b/src/jit_compiler_x86_static.asm
@@ -78,7 +78,7 @@ randomx_program_prologue PROC
 	and eax, RANDOMX_SCRATCHPAD_MASK
 	ror rdx, 32
 	and edx, RANDOMX_SCRATCHPAD_MASK
-	jmp randomx_program_loop_begin
+	jmp rx_program_loop_begin
 randomx_program_prologue ENDP
 
 ALIGN 64
@@ -86,6 +86,7 @@ ALIGN 64
 
 ALIGN 64
 randomx_program_loop_begin PROC
+rx_program_loop_begin::
 	nop
 randomx_program_loop_begin ENDP
 
@@ -199,7 +200,7 @@ randomx_sshash_init PROC
 	xor r14, r8
 	mov r15, qword ptr [r7_add]
 	xor r15, r8
-	jmp randomx_program_end
+	jmp rx_program_end
 randomx_sshash_init ENDP
 
 ALIGN 64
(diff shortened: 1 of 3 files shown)


src/jit_compiler_x86_static.asm
```diff
@@ -78,14 +78,15 @@ randomx_program_prologue PROC
...
-	jmp randomx_program_loop_begin
+	jmp rx_program_loop_begin
 randomx_program_prologue ENDP
 
 ALIGN 64
 	include asm/program_xmm_constants.inc
 
 ALIGN 64
 randomx_program_loop_begin PROC
+rx_program_loop_begin:
```
tevador, line 89: Should be a double colon here, otherwise the label is scoped to the procedure and the jmp instruction won't build.

src/jit_compiler_x86_static.asm
```diff
@@ -199,14 +200,15 @@ randomx_sshash_init PROC
...
-	jmp randomx_program_end
+	jmp rx_program_end
 randomx_sshash_init ENDP
 
 ALIGN 64
 	include asm/program_sshash_constants.inc
 
 ALIGN 64
 randomx_program_end PROC
+rx_program_end:
```
tevador, line 211: Same as above, double colon needed here.

tevador (changes requested):

src/jit_compiler_x86_static.asm
hyc: ok.

hyc (commented):
Also wondering if this needs similar treatment or not. Since it's just referring within the same procedure I guessed it was OK as-is. https://github.com/tevador/RandomX/pull/222/files#diff-4ba230c81224fe7330a5fbbde21150d765680cda2b379eeee43d41c5a9acd7d5R140

SChernykh (approved):

db4rTo:
> Use local symbols instead of global symbols, to prevent problems
> with dynamic linker.
> 
> Was causing reproducible crashes in monero core_tests.
> The .S patches have been tested on ARM and x86-64 Linux. The corresponding .asm patches are untested and need review, perhaps they're not needed.
> 
> The crashes were only in debug builds; they use dynamically linked libraries. This bug only hits when librandomx is a shared library.

Issue #216, opened 2021-07-23
Benchmark fails to run on 2012 Macbook in v1.1.9 (works in v1.1.8)


cryptonote-social:
Machine details:

>   Model Name:	MacBook Pro
>   Model Identifier:	  Model Name:	MacBook Pro
>   Model Identifier:	MacBookPro8,2
>   Processor Name:	Intel Core i7
>   Processor Speed:	2.2 GHz
>   Number of Processors:	1
>   Total Number of Cores:	4
>   L2 Cache (per Core):	256 KB
>   L3 Cache:	6 MB
>   Memory:	4 GB
>   Boot ROM Version:	87.0.0.0.0
>   SMC Version (system):	1.69f4MacBookPro8,2
>   Processor Name:	Intel Core i7
>   Processor Speed:	2.2 GHz
>   Number of Processors:	1
>   Total Number of Cores:	4
>   L2 Cache (per Core):	256 KB
>   L3 Cache:	6 MB
>   Memory:	4 GB
>   Boot ROM Version:	87.0.0.0.0
>   SMC Version (system):	1.69f4

Steps to reproduce:

> 1233][oldlaptop][~/src/github.com/tevador/RandomX/build]: git checkout v1.1.9
> HEAD is now at fe4324e Merge pull request #213 from hyc/m1typo
> [1234][oldlaptop][~/src/github.com/tevador/RandomX/build]: make -j 3
> [  8%] Building C object CMakeFiles/randomx.dir/src/argon2_ssse3.c.o
> ...
> [100%] Linking CXX executable randomx-tests
> [100%] Built target randomx-tests
> [1235][oldlaptop][~/src/github.com/tevador/RandomX/build]: ./randomx-benchmark --mine --jit --auto
> RandomX benchmark v1.1.8
>  - Argon2 implementation: SSSE3
>  - full memory mode (2080 MiB)
>  - JIT compiled mode 
>  - hardware AES mode
>  - small pages mode
>  - batch mode
> Initializing (8 threads) ...
> Memory initialized in 14.4383 s
> Initializing 1 virtual machine(s) ...
> ERROR: Cannot create VM with the selected options. Try using --softAes

Note that if I use v1.1.8 everything is fine:

> [1236][oldlaptop][~/src/github.com/tevador/RandomX/build]: git checkout v1.1.8
> Previous HEAD position was fe4324e Merge pull request #213 from hyc/m1typo
> HEAD is now at 5ce5f49 add --noBatch benchmark option
[...]

PR #226, opened 2021-11-12
Fix W^X policy for apple portability

jtgrassie:
`pthread_jit_write_protect_np` undefined on at least Mojave (10.14.6, x86_64).

### DIFF
--- a/src/virtual_memory.cpp
+++ b/src/virtual_memory.cpp
@@ -142,6 +142,8 @@ void setPagesRW(void* ptr, std::size_t bytes) {
 	&& MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
 	if (__builtin_available(macOS 11.0, *)) {
 		pthread_jit_write_protect_np(false);
+	} else {
+		pageProtect(ptr, bytes, PAGE_READWRITE);
 	}
 #else
 	pageProtect(ptr, bytes, PAGE_READWRITE);
@@ -153,6 +155,8 @@ void setPagesRX(void* ptr, std::size_t bytes) {
 	&& MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
 	if (__builtin_available(macOS 11.0, *)) {
 		pthread_jit_write_protect_np(true);
+	} else {
+		pageProtect(ptr, bytes, PAGE_EXECUTE_READ);
 	}
 #else
 	pageProtect(ptr, bytes, PAGE_EXECUTE_READ);
--- a/src/virtual_memory.cpp
+++ b/src/virtual_memory.cpp
@@ -36,12 +36,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifdef __APPLE__
 #include <mach/vm_statistics.h>
 #include <TargetConditionals.h>
+#include <AvailabilityMacros.h>
 # if TARGET_OS_OSX
-#  if TARGET_CPU_ARM64
-#   define USE_PTHREAD_JIT_WP	1
-#  else
-#   undef USE_PTHREAD_JIT_WP
-#  endif
+#  define USE_PTHREAD_JIT_WP	1
 #  include <pthread.h>
 # endif
 #endif
@@ -118,8 +115,11 @@ void* allocMemoryPages(std::size_t bytes) {
 	mem = mmap(nullptr, bytes, PAGE_READWRITE | RESERVED_FLAGS | PEXTRA, MAP_ANONYMOUS | MAP_PRIVATE | MEXTRA, -1, 0);
 	if (mem == MAP_FAILED)
 		throw std::runtime_error("allocMemoryPages - mmap failed");
-#ifdef USE_PTHREAD_JIT_WP
-	pthread_jit_write_protect_np(false);
+#if defined(USE_PTHREAD_JIT_WP) && defined(MAC_OS_VERSION_11_0) \
+	&& MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
+	if (__builtin_available(macOS 11.0, *)) {
+		pthread_jit_write_protect_np(false);
+	}
 #endif
 #endif
 	return mem;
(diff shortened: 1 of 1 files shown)


hyc:
@selsta tested the current code successfully on x86-64 already. If this is not working on Macos 10, then the ifdef should be against the OS version, not the CPU type. And if we need a single binary to work on multiple OS revisions, then we need a runtime check for this function's usability.

The patch in this PR should not be accepted as-is.

jtgrassie:
> @selsta tested the current code successfully on x86-64 already.

But did he test o earlier versions of macOS (i.e. before macOS 11 Big Sur)?

jtgrassie:
> IMO we should probably drop support for MacOS 10. 

Last macOS 10 (10.15) was released end of 2019, which is fairly recent.

selsta:
We cross compile with the 10.11 SDK in gitian / depends and I did not see any build failure yet.

So I'm surprised that it fails on 10.14

jtgrassie:
@selsta Monero build output against 10.15 SDK:

```
[ 40%] Building CXX object external/randomx/CMakeFiles/randomx.dir/src/virtual_memory.cpp.o
/Users/dog/monero/external/randomx/src/virtual_memory.cpp:118:2: error: use of undeclared identifier 'pthread_jit_write_protect_np'
        pthread_jit_write_protect_np(false);
        ^
/Users/dog/monero/external/randomx/src/virtual_memory.cpp:138:2: error: use of undeclared identifier 'pthread_jit_write_protect_np'
        pthread_jit_write_protect_np(false);
        ^
/Users/dog/monero/external/randomx/src/virtual_memory.cpp:146:2: error: use of undeclared identifier 'pthread_jit_write_protect_np'
        pthread_jit_write_protect_np(true);
```

selsta:
With this patch `MAP_JIT` doesn't get set anymore on x86-64 so that has to be changed.

selsta:
An `ifdef` for the OS version (11 or newer) would be fine too. I think not using `pthread_jit_write_protect_np` on older Intel Macs shouldn't matter.

jtgrassie:
> An ifdef for the OS version (11 or newer) would be fine too. I think not using pthread_jit_write_protect_np on older Intel Macs shouldn't matter.

Which goes back to @hyc's point:

>> And if we need a single binary to work on multiple OS revisions, then we need a runtime check for this function's usability.

As it stands, this patch allows to build for target ARM **or** X86_64. If we want to a single binary that will work on _both_, we need runtime checks for `pthread_jit_write_protect_np`. I can rework with the runtime checks if desired, the only reason I didn't add was because of the unknown overhead.

jtgrassie:
@selsta df45079 with runtime checking.

selsta:
> If we want to a single binary that will work on both, we need runtime checks for pthread_jit_write_protect_np

Can you explain why? We can use `lipo` to combine an Intel binary compiled with 10.11 SDK and an ARM binary compiled with 11 SDK. No runtime checks required, but maybe I'm missing something.

jtgrassie:
> Can you explain why?

It's not really an architecture issue but rather an SDK issue. `pthread_jit_write_protect_np` simply doesn't exist on < MacOS 11, hence the undefined errors.

selsta:
> pthread_jit_write_protect_np simply doesn't exist on < MacOS 11

That's why I suggested to simply ifdef for the macOS version. Basically your current patch without the `__builtin_available ` runtime check. But like I said, maybe I'm missing something.

jtgrassie:
The `__builtin_available` is for the _runtime_ checks. Hence this will run on 10 and 11.

selsta:
do you need `AvailabilityMacros.h`?

selsta:
and I still don't understand why you need a runtime check but I will let @hyc review this

selsta:
The weird thing is CI compiles RandomX 1.1.10 under macOS 10.15 without any build errors:

https://github.com/monero-project/monero/runs/4144066544?check_suite_focus=true

jtgrassie:
> and I still don't understand why you need a runtime check

My understanding is that `pthread_jit_write_protect_np` is only _implemented_ on apple silicon. So you could be running on 11 but not ARM. I don't have an apple silicon machine to test (though your lipo suggestion should also work), but without this patch, it's not possible to even compile or run on < 11.

jtgrassie:
> The weird thing is CI compiles RandomX 1.1.10 under macOS 10.15 without any build errors

Because it's using `MacOSX11.1.sdk`.

selsta:
>Because it's using MacOSX11.1.sdk.

True, seems Github updates Xcode in the VM image. Only question remaining is why cross compiling still worked with the 10.11 SDK.

jtgrassie:
> Only question remaining for me is why cross compiling still worked with the 10.11 SDK

It only "worked" insofar as it compiled something that won't run on 10.x.

src/virtual_memory.cpp
```diff
@@ -134,16 +138,22 @@ static inline void pageProtect(void* ptr, std::size_t bytes, int rules) {
 }
 
 void setPagesRW(void* ptr, std::size_t bytes) {
-#ifdef USE_PTHREAD_JIT_WP
-	pthread_jit_write_protect_np(false);
+#if defined(USE_PTHREAD_JIT_WP) && defined(MAC_OS_VERSION_11_0) \
+	&& MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
+	if (__builtin_available(macOS 11.0, *)) {
+		pthread_jit_write_protect_np(false);
+	}
```
hyc, line 145: This is a no-op if `pthread_jit_write_protect_np` is not available. If it's not available it should be using `pageProtect`.

That was the entire point of issue #216. Let's not run around in circles here.

hyc:
> > Only question remaining for me is why cross compiling still worked with the 10.11 SDK
> 
> It only "worked" insofar as it compiled something that won't run on 10.x.

Still, it should not have linked successfully then, when built with the 10.x SDK. Something still doesn't make sense here.

jtgrassie:
> Still, it should not have linked successfully then, when built with the 10.x SDK. Something still doesn't make sense here.

It wasn't built with the 10.x SDK. It won't build with the 10.x SDK. That's the what started me down this rabbit hole.

selsta:
gitian cross compiling was built with the 10.11 SDK.

https://github.com/monero-project/monero/runs/4144066674?check_suite_focus=true

jtgrassie:
> gitian cross compiling was built with the 10.11 SDK.
> 
> https://github.com/monero-project/monero/runs/4144066674?check_suite_focus=true

That is a different matter (gitian), because `pthread_jit_write_protect_np` _does not exist_ on 10.x. This must mean `USE_PTHREAD_JIT_WP` is not getting defined in that gitian build (if it was, you'd get an undefined on `pthread_jit_write_protect_np`). 

If you want to build on 10 or 11 to create a single binary that can run on 10 & 11 (using `pthread_jit_write_protect_np` when available), you need this PR.

hyc:
I've run this patch on my M1 Macbook Pro with no apparent impact on randomx-benchmark performance or correctness. LGTM.

hyc (approved):

hyc:
Final thought - if there is no performance gain on x86 for using `pthread_jit_write_protect_np` then maybe we should just go back to the earlier commit where it was only used on aarch64.

selsta:
> Final thought - if there is no performance gain on x86 for using pthread_jit_write_protect_np then maybe we should just go back to the earlier commit where it was only used on aarch64.

Might be the easiest solution, yes.

jtgrassie:
> Final thought - if there is no performance gain on x86 for using `pthread_jit_write_protect_np` then maybe we should just go back to the earlier commit where it was only used on aarch64.

Well, this has all the compile _and_ runtime checks now, so is more portable.

selsta (approved):
Tested on:

- Intel 10.12.6, 10.15.7, 11.6.1
- ARM 12.1

PR #231, opened 2021-12-17
Pr nyuu a64 c1

Nyuu:
Hi,

while trying to understand the code and working on it, i noticed that in the 'src/jit_compiler_a64.hpp ' file, my markup did not work properly because according to KDevelop, the were several uses of undeclared 'class JitCompilerA64' references.

I just added the code at line 43 next to the other declarations (as it is done in the 'src/jit_compiler_x86.hpp') and the markup works now and i get no errors in my IDE anymore.
Not sure if this really is a code error or just a very sensitive IDE as i am new to KDevelop ^^"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ x64/
 Release/
 Debug/
 build/
+.kdev4
+RandomX-tk1.pbg.su.kdev4
+sekrit/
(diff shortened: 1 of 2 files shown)


Nyuu:
Ouch ^^"
Sorry, that was a nobrainer and realised i still had the pr open.

PR #240, opened 2022-05-07
Fix use of __builtin_available() on MacOSX

hyc:
It's not actually built in, causes linking errors.
Unfortunately not using it trips warnings, but they're harmless

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,6 +181,14 @@ if(TARGET generate-asm)
   add_dependencies(randomx generate-asm)
 endif()
 
+if(APPLE)
+# Change symbol ___isOSVersionAtLeast from global to static. See comment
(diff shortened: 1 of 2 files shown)


hyc:
Also, the version check is a hack. it will be correct for MacOSX 11.0 and up; it will report lower-than-actual numbers for 10.15 and older. For our purpose, that's not a problem.

hyc:
To be clear: it causes linking errors when cross-compiling from Linux, using a Mac SDK of 11.0 or newer. On older SDKs the compiler knows that OS 11.0 isn't supported by that SDK and just no-ops the relevant code, which is why there's no link error in that case.

hyc:
New version of patch works with clang's __builtin_available(), to avoid pointless warnings.

PR #248, opened 2022-06-04
Python package

xloem:
Hey,

I wrote this basic cython wrapper for the library.  It seems to work for me. 

I ran into some trouble figuring out how to set up a build system for distribution, but I think it's still valuable.

```
$ cd python
$ pip3 install .
```
```
$ python3
>>> import randomx
>>> vm = randomx.RandomX(b'key material', full_mem=True)
>>> hash = vm(b'message')
>>> print(hash)
>>> for hash in vm.calculate_hashes([b'message one', b'message two', b'message three']):
...     print(hash)
...
```

EDIT:
What I ended up doing was copying the logic from CMakeLists.txt into setup.py . This lets this work: `pip3 install git+https://github.com/xloem/RandomX\#subdirectory=python` . To remove the embedded sources, revert to commit 292a506 .

### DIFF
--- a/python/crandomx.pxd
+++ b/python/crandomx.pxd
@@ -1,0 +1,53 @@
+cdef extern from "../src/randomx.h":
+
+    ctypedef enum randomx_flags:
+        RANDOMX_FLAG_DEFAULT
+        RANDOMX_FLAG_LARGE_PAGES
+        RANDOMX_FLAG_HARD_AES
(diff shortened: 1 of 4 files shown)


tevador:
I think this should be in its own repository, referencing RandomX as a submodule.

xloem:
Thanks for your reply :)

PR #257, opened 2022-10-31
Fix compilation and JIT support on aarch64_be

bkuhls:
Fixes building monero & p2pool with buildroot toolchains.


tevador:
The current implementation of JIT won't work on big endian processors. This PR doesn't actually fix anything, it only enables JIT that will give wrong hashes (or possibly crash).

bkuhls:
Thanks for the info and sorry for the noise ;)

PR #259, opened 2022-11-30
Invalidate CPU I-cache on Apple Silicon

SChernykh:
Call sys_icache_invalidate() after pthread_jit_write_protect_np() according to https://developer.apple.com/documentation/apple-silicon/porting-just-in-time-compilers-to-apple-silicon

### DIFF
--- a/src/virtual_memory.c
+++ b/src/virtual_memory.c
@@ -189,6 +189,7 @@ void setPagesRX(void* ptr, size_t bytes) {
 	&& MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
 	if (__builtin_available(macOS 11.0, *)) {
 		pthread_jit_write_protect_np(1);
+		__builtin___clear_cache((char*)ptr, ((char*)ptr) + bytes);
 	} else {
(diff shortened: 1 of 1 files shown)


hyc:
This issue keeps coming up a lot. Just a reminder (for future reference): we're already using the gcc/clang __builtin___clear_cache() and on MacOS that is simply a call to sys_icache_invalidate(). So the real question to answer is, do we have enough of those __builtin___clear_cache() invocations, in the right places?

src/virtual_memory.c
```diff
@@ -39,6 +39,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #  include <pthread.h>
 #  include <sys/utsname.h>
 #  include <stdio.h>
+#  if defined(MAC_OS_VERSION_11_0) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_VERSION_11_0
```
hyc, line 42: I think we should drop this chunk and just use `__builtin___clear_cache()` like jit_compiler_a64.cpp does.

hyc (approved):
Approving, even though it seems this isn't fixing any crashes. It is at least following the letter of the Apple docs.

PR #265, opened 2023-04-22
RandomX commitments with double-hashing

tevador:
Related to this proposal: https://github.com/monero-project/monero/issues/8827

The commitment is calculated from the hash and its input (i.e. the block header). The idea is that the output of `randomx_calculate_hash` will be appended to the block header and the difficulty will be calculated based on the output of `randomx_calculate_commitment`, which is very fast.


src/randomx.cpp
```diff
@@ -400,4 +400,15 @@ extern "C" {
...
 	}
+
+	void randomx_calculate_hash_v2(const void* input, size_t inputSize, const void* v1_in, void* v2_out) {
+		assert(inputSize == 0 || input != nullptr);
+		assert(v1_in != nullptr);
+		assert(v2_out != nullptr);
+		blake2b_state state;
+		blake2b_init(&state, RANDOMX_HASH_SIZE);
+		blake2b_update(&state, input, inputSize);
+		blake2b_update(&state, v1_in, RANDOMX_HASH_SIZE);
```
SChernykh, line 411: Can this be a single blake2b_update call in optimized code?

tevador: Yes, it should be possible, assuming the V1 hash immediately follows the hashing blob in memory. However, the savings will be quite small, perhaps a few nanoseconds per hash, because `blake2b_update` doesn't call `blake2b_compress` unless the accumulated data exceeds 128 bytes.

SChernykh (approved):

tevador:
I'm not planning any modifications to this PR unless something comes up during the review. The changes are fairly small, so it's mostly about the concept.

One thing to note is that batched hashing using the `randomx_calculate_hash_next` function will become slightly more complicated due to the need to recall the previous hash input when the v1 result is ready.

SChernykh:
> One thing to note is that batched hashing using the `randomx_calculate_hash_next` function will become slightly more complicated due to the need to recall the previous hash input when the v1 result is ready.

That wouldn't be an issue for mining software because it only needs to remember the previous nonce value (4 bytes). But yes, `randomx_calculate_hash_next` will need two more parameters for the previous input.

src/randomx.cpp
```diff
@@ -400,4 +400,15 @@ extern "C" {
 		machine->run(machine->tempHash);
 		machine->getFinalResult(output, RANDOMX_HASH_SIZE);
 	}
+
+	void randomx_calculate_hash_v2(const void* input, size_t inputSize, const void* v1_in, void* v2_out) {
+		assert(inputSize == 0 || input != nullptr);
+		assert(v1_in != nullptr);
+		assert(v2_out != nullptr);
```
bitcartel, line 407: Hi, this can help prevent user error when calling the library.
assert(v1_in != v2_out);

tevador: It will work perfectly fine with `v1_in == v2_out`.

tevador:
I renamed the new function to `randomx_calculate_commitment` because it provides a commitment to a RandomX hash value. The "v2" naming was not a good choice because this change is orthogonal to future VM changes that might need to be versioned.

SChernykh (approved):

Issue #270, opened 2023-07-29
Segfault on sample code on init cache


FreeTrade:
Hoping for some pointers on what I might be doing wrong here -

Got the following sample code -

```
    const char myKey[] = "RandomX example key";
    const char myInput[] = "RandomX example input";
    char hash[RANDOMX_HASH_SIZE];
    std::cout << "Get flags" << std::endl;
    randomx_flags flags = randomx_get_flags();
    std::cout << flags << std::endl;
    std::cout << "Allocate cache" << std::endl;
    randomx_cache *myCache = randomx_alloc_cache(flags);
	if (myCache == nullptr) {
		std::cout << "Cache allocation failed" << std::endl;
		return SerializeHash(*this);
	}
	std::cout << "Init cache" << std::endl;
	randomx_init_cache(myCache, &myKey, sizeof myKey);
    std::cout << "MyMachine" << std::endl;
```

with output -

```
Get flags
106
Allocate cache
Init cache
Segmentation fault (core dumped)
```

free -h is
```             
                    total        used        free      shared  buff/cache   available
Mem:           3.8Gi       262Mi       847Mi       1.0Mi       2.7Gi       3.3Gi
Swap:          8.0Gi        40Mi       8.0Gi
```

Running ubuntu 23.04.
Tried compiling with both -DARCH=native and no -DARCH

tevador:
Which CPU are you using?

Can you provide a stack trace? Run a debug build with `gdb` and when it crashes, use the `bt` command.

FreeTrade:
Thanks for the help, here's that info
```
Get flags
0
Allocate cache

Program received signal SIGSEGV, Segmentation fault.
0x0000555555a81a37 in randomx::generateSuperscalar(randomx::SuperscalarProgram&, randomx::Blake2Generator&) ()
(gdb) bt
#0  0x0000555555a81a37 in randomx::generateSuperscalar(randomx::SuperscalarProgram&, randomx::Blake2Generator&) ()
#1  0x0000555555a7f72d in randomx::initCache(randomx_cache*, void const*, unsigned long) ()
#2  0x0000555555a75067 in randomx_init_cache ()
#3  0x00005555559df520 in CBlockHeader::GetHash (this=this@entry=0x555555dc2098 <mainParams+1848>)
    at primitives/block.cpp:38
#4  0x00005555558e002d in CMainParams::CMainParams (this=0x555555dc1960 <mainParams>) at chainparams.cpp:174
[...]

PR #272, opened 2023-08-14
Update superscalar.cpp

FreeTrade:
Received this suggestion from @anshulmttl to address problem I was having at

https://github.com/tevador/RandomX/issues/270

It seems to compile and run fine, although I'm unsure if it could have any side effects.

Comment ...

The static members were not being initialized so i have initialized them properly now.  

I found that SupercalarInstruction::Null is not being initialized whereas the SuperscalarInstructionInfo::NOP is initialized just fine.

So i have made changed to SuperscalarInstruction to initialize SuperscalarInstruction::Null properly

### DIFF
--- a/src/superscalar.cpp
+++ b/src/superscalar.cpp
@@ -203,9 +203,10 @@ namespace randomx {
 		int resultOp_ = 0;
 		int dstOp_ = 0;
 		int srcOp_;
-
+	public:
 		SuperscalarInstructionInfo(const char* name)
-			: name_(name), type_(SuperscalarInstructionType::INVALID), latency_(0) {}
+			: name_(name), type_(SuperscalarInstructionType::INVALID), latency_(0) {
+			}
 		SuperscalarInstructionInfo(const char* name, SuperscalarInstructionType type, const MacroOp& op, int srcOp)
 			: name_(name), type_(type), latency_(op.getLatency()), srcOp_(srcOp) {
 			ops_.push_back(MacroOp(op));
@@ -572,8 +573,17 @@ namespace randomx {
 			src_ = dst_ = -1;
 			canReuse_ = groupParIsSource_ = false;
 		}
-
 		SuperscalarInstruction(const SuperscalarInstructionInfo* info) : info_(info) {
+		}
+		
+	public:
+		SuperscalarInstruction(const SuperscalarInstruction& obj1) : info_(obj1.info_)
+		{
+			if(!obj1.info_)
+			{
+				info_ = (&SuperscalarInstructionInfo::NOP);
+			}
+
 		}
 	};
 
--- a/src/superscalar.cpp
+++ b/src/superscalar.cpp
@@ -203,7 +203,7 @@ namespace randomx {
 		int resultOp_ = 0;
 		int dstOp_ = 0;
 		int srcOp_;
-	
+	//public:
 		SuperscalarInstructionInfo(const char* name)
 			: name_(name), type_(SuperscalarInstructionType::INVALID), latency_(0) {
 			}
@@ -579,10 +579,20 @@ namespace randomx {
 	public:
 		SuperscalarInstruction(const SuperscalarInstruction& obj1) : info_(obj1.info_)
 		{
+			
 			if(!obj1.info_)
 			{
 				info_ = (&SuperscalarInstructionInfo::NOP);
 			}
+			//else
+			src_ = obj1.src_;
+			dst_ = obj1.dst_;
+			mod_ = obj1.mod_;
+			imm32_ = obj1.imm32_;
+			opGroup_ = obj1.opGroup_;
+			opGroupPar_ = obj1.opGroupPar_;
+			canReuse_ = obj1.canReuse_;
+			groupParIsSource_ = obj1.groupParIsSource_;
 
 		}
 	};
--- a/src/superscalar.cpp
+++ b/src/superscalar.cpp
@@ -203,7 +203,7 @@ namespace randomx {
 		int resultOp_ = 0;
 		int dstOp_ = 0;
 		int srcOp_;
-	public:
+	
 		SuperscalarInstructionInfo(const char* name)
 			: name_(name), type_(SuperscalarInstructionType::INVALID), latency_(0) {
 			}


SChernykh:
> I found that SupercalarInstruction::Null is not being initialized whereas the SuperscalarInstructionInfo::NOP is initialized just fine.

According to C++ standard, both must be initialized properly, so you're fixing non-existing bug which is caused by your incorrect compilation setup. You should fix your CMake files instead.

src/superscalar.cpp
```diff
@@ -203,9 +203,10 @@ namespace randomx {
 		int resultOp_ = 0;
 		int dstOp_ = 0;
 		int srcOp_;
-
+	public:
```
anshulmttl, line 206: This can be removed. I had added it only for testing.

src/superscalar.cpp
```diff
@@ -572,9 +573,18 @@ namespace randomx {
 			src_ = dst_ = -1;
 			canReuse_ = groupParIsSource_ = false;
 		}
-
 		SuperscalarInstruction(const SuperscalarInstructionInfo* info) : info_(info) {
 		}
+		
+	public:
+		SuperscalarInstruction(const SuperscalarInstruction& obj1) : info_(obj1.info_)
```
SChernykh, line 580: First of all, see my comment in the conversation - you're fixing a non-existent bug. Second, if you're adding a copy constructor, you must copy all class fields, not just `info_`. This PR is unlikely to be merged, I'm just pointing out errors in implementation.

FreeTrade: Thanks for your review. I note your comment that the fix should be in a different place, but resolving the issue of copying all class fields.

anshulmttl: As i see in your code the SuperscalarInstruction copy constructor is called only 1 location in **void generateSuperscalar(SuperscalarProgram& prog, Blake2Generator& gen)** and that call is **SuperscalarInstruction currentInstruction = SuperscalarInstruction::Null;**. So even if i don't copy all fields it will work same as copying all fields in current execution of BCHUnlimited. But keeping coding standards in mind i feel it is better to copy all fields and have made the changes.

anshulmttl: For your 1st comment, in software engineering we don't define a bug as non existent. A bug is found only after it has been tested in various environments and not just unit test cases designed to work. So BCHUnlimited is clearly a new environment for your software.

anshulmttl:
> > I found that SupercalarInstruction::Null is not being initialized whereas the SuperscalarInstructionInfo::NOP is initialized just fine.
> 
> According to C++ standard, both must be initialized properly, so you're fixing non-existing bug which is caused by your incorrect compilation setup. You should fix your CMake files instead.

In software engineering we don't define a bug as non existent. A bug is found only after it has been tested in various environments and not just unit test cases designed to work. So BCHUnlimited is clearly a new environment for your software.

SChernykh:
Of course the bug exists, in  BCHUnlimited makefiles. It's just not in this cpp file. You're fixing a symptom, not the actual problem. How can you know that nothing else is broken because of that? So this why I'm saying that this PR will not be merged.

anshulmttl:
> 

As per C++ standard the classes should be declared in header files and defined in C++ files but your code does not follow coding standards.

anshulmttl:
> Of course the bug exists, in BCHUnlimited makefiles. It's just not in this cpp file. You're fixing a symptom, not the actual problem. How can you know that nothing else is broken because of that? So this why I'm saying that this PR will not be merged.
> 
> P.S. I'll repeat again - according to C++ standard, your fix will do nothing because `info_` will never be `nullptr` here

Your constructor is private so as per C++ standard your class cannot have any instance of SuperscalarInstruction except for SuperscalarInstruction::Null or info_ = SuperscalarInstructionInfo::NOP. info_ cannot be null ptr and infact you cannot even initialize class SuperscalarInstruction because it is private constructor. Check your C++ concepts.

SChernykh:
> As per C++ standard the classes should be declared in header files and defined in C++ files but your code does not follow coding standards.

Coding standards are not C++ standard. If this code broke C++ standard, it wouldn't compile at all.

> info_ cannot be null ptr and infact you cannot even initialize class SuperscalarInstruction because it is private constructor. Check your C++ concepts.

So? Why do you check for nullptr in this PR then? Don't you see a contradiction here?

tevador:
> As per C++ standard the classes should be declared in header files and defined in C++ files but your code does not follow coding standards.

Can you link to the relevant section of the C++ standard?

This PR won't be merged unless you can show that the code actually violates the specs. The fact that your code crashes is not a proof. There can be other reasons for that such as: a compiler bug or an incorrect build setup. This code has been [audited](https://github.com/tevador/RandomX#audits) and used for 4 years without issues on various platfoms and nobody has reported a problem.

> I'm unsure if it could have any side effects.

A "side effect" could cause unintentional hard forks in the cryptocurrencies that use RandomX (and there are many), so any attempts to fix bugs must be done very carefully in this repository.

SChernykh:
Even if there is a bug in this repository, it should be most likely fixed in `CMakeLists.txt`, and not in this .cpp file. But I give 99% chance it's something wrong with BCHUnlimited makefiles, or the way you try to integrate RandomX there.

anshulmttl:
> > As per C++ standard the classes should be declared in header files and defined in C++ files but your code does not follow coding standards.

It is a general concept. 

You have used inline functions in your code which is not good as per c++ standard. You have declared class in cpp file in instead of header file which is not good as per coding standards. Searching on google for these 2 will give you ample amount of information. Just for reference here is first reference i found.
https://www.learncpp.com/cpp-tutorial/class-code-and-header-files/

SChernykh:
Based on your comments, you don't even understand what is "C++ standard" and what is "Coding guide". Please don't try to teach us here while you yourself forgot how to write a proper copy constructor.

tevador:
> It is a general concept.

Header files declare public interfaces that are accessible across translation units. The class `SuperscalarInstruction` is an implementation detail of the public `generateSuperscalar` function and it not meant to be visible outside of superscalar.cpp.

SChernykh:
RandomX uses C++14: https://www.iso.org/standard/64029.html - this is THE standard I was talking about. Not some random articles on learncpp.

anshulmttl:
> Based on your comments, you don't even understand what is "C++ standard" and what is "Coding guide". Please don't try to teach us here while you yourself forgot how to write a proper copy constructor.

See it is written in 1st line of google why not to use inline functions :)
https://www.google.com/search?q=issues+with+using+inline+functions&sca_esv=556671154&sxsrf=AB5stBhFiuZqPNrzPM3OncLxS3O41EBGaQ%3A1691998625085&ei=odnZZN3jBIOjseMPmt6YuAQ&ved=0ahUKEwiduIiI0tuAAxWDUWwGHRovBkcQ4dUDCBA&uact=5&oq=issues+with+using+inline+functions&gs_lp=Egxnd3Mtd2l6LXNlcnAiImlzc3VlcyB3aXRoIHVzaW5nIGlubGluZSBmdW5jdGlvbnMyBRAhGKABMggQIRgWGB4YHTIIECEYFhgeGB1IzSxQiAdYvitwA3gBkAEAmAGmAaAB3R6qAQQ0LjMwuAEDyAEA-AEBwgIKEAAYRxjWBBiwA8ICBxAjGIoFGCfCAgQQIxgnwgIIEAAYigUYkQLCAgsQABiABBixAxiDAcICBRAuGIAEwgILEC4YigUYsQMYgwHCAhEQLhiABBixAxiDARjHARjRA8ICERAuGIAEGLEDGIMBGMcBGK8BwgIKEAAYigUYsQMYQ8ICBxAAGIoFGEPCAgcQLhiKBRhDwgINEC4YigUYsQMYgwEYQ8ICBRAAGIAEwgIOEC4YgAQYsQMYxwEY0QPCAggQABiABBixA8ICChAAGIAEGEYY_wHCAgoQABiABBgUGIcCwgIGEAAYFhgewgIHECEYoAEYCsICBBAhGBXCAgoQIRgWGB4YDxgd4gMEGAAgQYgGAZAGCA&sclient=gws-wiz-serp

And see it is written on google first line why to declare class in header files. 
https://www.google.com/search?q=why+to+declare+class+in+header+file&sca_esv=556671154&sxsrf=AB5stBhnkIvBLzyL2D6s3DBoMyPsNuNC-A%3A1691998644873&ei=tNnZZP3qNNSgseMPhYCHqAo&ved=0ahUKEwj9lMCR0tuAAxVUUGwGHQXAAaUQ4dUDCBA&uact=5&oq=why+to+declare+class+in+header+file&gs_lp=Egxnd3Mtd2l6LXNlcnAiI3doeSB0byBkZWNsYXJlIGNsYXNzIGluIGhlYWRlciBmaWxlMgcQIRigARgKMgcQIRigARgKMggQIRgWGB4YHUiiNVDFBVizNHADeAGQAQCYAYcCoAGzIqoBBjcuMjUuMrgBA8gBAPgBAcICChAAGEcY1gQYsAPCAgcQIxiKBRgnwgINEC4YigUYsQMYgwEYQ8ICCBAAGIoFGJECwgINEAAYigUYsQMYgwEYQ8ICCxAAGIAEGLEDGIMBwgIEECMYJ8ICBxAAGIoFGEPCAggQLhiABBixA8ICCxAAGIoFGLEDGIMBwgILEC4YgAQYsQMYgwHCAgUQABiABMICCxAuGIoFGLEDGIMBwgIFEC4YgATCAgoQABiKBRixAxhDwgIIEAAYgAQYsQPCAgoQABiABBgUGIcCwgIGEAAYFhgewgIIEAAYFhgeGA_CAggQABgIGB4YDcICCBAAGIoFGIYDwgIFECEYoAHCAgoQIRgWGB4YDxgdwgIEECEYFeIDBBgAIEGIBgGQBgg&sclient=gws-wiz-serp

anshulmttl:
I don't want to fight on this context again and again. I have just sent you the text which is available at google. I don't want any further discussion on this. You can decide on your own after reading this google articles.

anshulmttl:
Please check tevador's reply. He has given the correct reply to my point. It is your design consideration. coding standard in C++ says it is better to write class declaration in header files, it does not say that class definitions cannot be written in CPP files or inside functions.

>

SChernykh:
And my point is that this whole PR and conversation shouldn't have existed at all. The bug is somewhere else. Also, your point is not related to this PR at all, we're here not to discuss best coding practices, but to discuss one specific issue that @FreeTrade has.

FreeTrade:
Note: On further inspection - while this change allowed my code to compile and run in the build system, it only worked with very few flags set and produced incorrect hashes. Recommend against using.

PR #274, opened 2023-09-08
RandomX v2 virtual machine changes

tevador:
* CFROUND becomes conditional with a 1/16 chance of writing into fprc
    * F and E registers are mixed together with AES instead of XOR

This PR is incomplete. Currently, only the X86 and portable versions work, hardware AES is needed and the changes are hardcoded. But it's enough to run some benchmarks.

* [x] CFROUND changes in the interpreter
* [x] CFROUND changes in X86 JIT compiler
* [ ] CFROUND changes in A64 JIT compiler
* [ ] V1/V2 selectable at runtime
* [x] New portable intrinsics
* [x] New X86 instrinsics
* [ ] New ARM64 instrinsics
* [ ] New PowerPC intrinsics
* [ ] Support for soft AES in JIT compiler

PR #275, opened 2023-10-08
JIT compiler for RISC-V

tevador:
Currently it assumes RV64GC as the baseline, which matches what Linux expects in terms of ISA extensions.

Additionally, when configured with `-DARCH=native`, it will check for the presence of two additional extensions:

* Zba - address generation (useful for the IADD_RS instruction)
* Zbb - basic bit manipulation (useful mainly for rotations)

These two extensions give about a 3-5% speed-up compared to the base ISA and a 10% speed-up for cache initialization (Argon2 is heavy on rotations).

Two additional extensions will be useful in the future: V (for SIMD) and Zkn (for AES), but AFAIK there are currently no chips that support them.

Here are some benchmarks with my StarFive JH7110 based board:

```
> ./randomx-benchmark --auto --mine --largePages --threads 4
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing (4 threads) ...
Memory initialized in 41.209 s
Initializing 4 virtual machine(s) ...
Calculated result: 10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Reference result:  10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Performance: 74.8368 hashes per second
```

The mining performance is about 75% of the Raspberry Pi 4, which is not bad considering the lack of vector instructions.

```
> ./randomx-benchmark --auto --verify --largePages
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - light memory mode (256 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing ...
Memory initialized in 4.30378 s
Calculated result: 10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Reference result:  10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Performance: 107.174 ms per hash
```

The verification performance is probably more useful as it's substantially faster than the interpreter, which takes over 1300 ms to verify a hash.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,42 @@ if(ARM_ID STREQUAL "aarch64" OR ARM_ID STREQUAL "arm64" OR ARM_ID STREQUAL "armv
   endif()
 endif()
 
+# RISC-V
+if(ARCH_ID STREQUAL "riscv64")
+  list(APPEND randomx_sources
+    src/jit_compiler_rv64_static.S
+    src/jit_compiler_rv64.cpp)
+  # cheat because cmake and ccache hate each other
+  set_property(SOURCE src/jit_compiler_rv64_static.S PROPERTY LANGUAGE C)
+  set_property(SOURCE src/jit_compiler_rv64_static.S PROPERTY XCODE_EXPLICIT_FILE_TYPE sourcecode.asm)
+
+  # default build uses the RV64GC baseline
+  set(RVARCH "rv64gc")
+
+  # for native builds, enable Zba and Zbb if supported by the CPU
+  if(ARCH STREQUAL "native")
+    enable_language(ASM)
+    try_run(RANDOMX_ZBA_RUN_FAIL
+        RANDOMX_ZBA_COMPILE_OK
+        ${CMAKE_CURRENT_BINARY_DIR}/
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/tests/riscv64_zba.s
+        COMPILE_DEFINITIONS "-march=rv64gc_zba")
+    if (RANDOMX_ZBA_COMPILE_OK AND NOT RANDOMX_ZBA_RUN_FAIL)
+      set(RVARCH "${RVARCH}_zba")
+    endif()
+    try_run(RANDOMX_ZBB_RUN_FAIL
+        RANDOMX_ZBB_COMPILE_OK
+        ${CMAKE_CURRENT_BINARY_DIR}/
+        ${CMAKE_CURRENT_SOURCE_DIR}/src/tests/riscv64_zbb.s
+        COMPILE_DEFINITIONS "-march=rv64gc_zbb")
+    if (RANDOMX_ZBB_COMPILE_OK AND NOT RANDOMX_ZBB_RUN_FAIL)
+      set(RVARCH "${RVARCH}_zbb")
+    endif()
+  endif()
+
+  add_flag("-march=${RVARCH}")
+endif()
+
 set(RANDOMX_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/src" CACHE STRING "RandomX Include path")
 
 add_library(randomx ${randomx_sources})
--- a/src/common.hpp
+++ b/src/common.hpp
@@ -116,12 +116,19 @@ namespace randomx {
 
 #if defined(_M_X64) || defined(__x86_64__)
 	#define RANDOMX_HAVE_COMPILER 1
+	#define RANDOMX_COMPILER_X86
 	class JitCompilerX86;
 	using JitCompiler = JitCompilerX86;
 #elif defined(__aarch64__)
 	#define RANDOMX_HAVE_COMPILER 1
+	#define RANDOMX_COMPILER_A64
 	class JitCompilerA64;
 	using JitCompiler = JitCompilerA64;
+#elif defined(__riscv) && __riscv_xlen == 64
+	#define RANDOMX_HAVE_COMPILER 1
+	#define RANDOMX_COMPILER_RV64
+	class JitCompilerRV64;
+	using JitCompiler = JitCompilerRV64;
 #else
 	#define RANDOMX_HAVE_COMPILER 0
 	class JitCompilerFallback;
--- a/src/jit_compiler_rv64.cpp
+++ b/src/jit_compiler_rv64.cpp
@@ -1,0 +1,1175 @@
+/*
+Copyright (c) 2023 tevador <tevador@gmail.com>
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+	* Redistributions of source code must retain the above copyright
+	  notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above copyright
+	  notice, this list of conditions and the following disclaimer in the
+	  documentation and/or other materials provided with the distribution.
+	* Neither the name of the copyright holder nor the
+	  names of its contributors may be used to endorse or promote products
+	  derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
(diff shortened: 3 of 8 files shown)


CMakeLists.txt
```diff
@@ -173,6 +173,42 @@ if(ARM_ID STREQUAL "aarch64" OR ARM_ID STREQUAL "arm64" OR ARM_ID STREQUAL "armv
...
 endif()
 
+# RISC-V
+if(ARCH_ID STREQUAL "riscv64")
+  list(APPEND randomx_sources
+    src/jit_compiler_rv64_static.S
+    src/jit_compiler_rv64.cpp)
+  # cheat because cmake and ccache hate each other
+  set_property(SOURCE src/jit_compiler_rv64_static.S PROPERTY LANGUAGE C)
+  set_property(SOURCE src/jit_compiler_rv64_static.S PROPERTY XCODE_EXPLICIT_FILE_TYPE sourcecode.asm)
```
SChernykh, line 183: I don't think there are any RISC-V Apple devices, so this Xcode check is unnecessary.

src/jit_compiler.hpp
```diff
@@ -28,10 +28,48 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #pragma once
 
-#if defined(_M_X64) || defined(__x86_64__)
+#include "common.hpp"
+
+namespace randomx {
+
+	struct CodeBuffer {
```
SChernykh, line 35: If this struct is only used by risc-v compiler, maybe it should be in `jit_compiler_rv64.hpp`

hyc:
I've got my Lichee Pi 4a now but haven't set it up yet. Will try this in a few days.

CMakeLists.txt
tevador: It doesn't hurt to leave it there so we are ready if (when) Apple decides to switch to RISC-V.

src/jit_compiler.hpp
tevador: This PR contains some seemingly unused code that will be used by #274. That includes this struct (will be used by the X86 compiler) and some assembly code in `jit_compiler_rv64_static.S` (e.g. the soft AES code).

src/tests/riscv64_zkn.s
```diff
@@ -0,0 +1,10 @@
+/* RISC-V - test if the Zkn extension is present */
```
SChernykh, line 1: This file is not referenced in CMakeLists.txt, and I don't see an `rx_aesenc_vec_i128` implementation. Is hardware AES not implemented yet?

SChernykh:
The code looks good. How many hashes did you run the benchmark for? I think it needs to be tested with at least 10M hashes and the result must be identical to what x64/aarch64 versions produce.

src/tests/riscv64_zkn.s
tevador: Yes, this PR does not include hardware AES. I removed the file for now because it's unlikely to be needed in the near future.

tevador:
I'm running 1M now (should take a couple hours) and then I'll try 10M. Then we'll have to repeat the runs with the native build. Do you have the correct hashes for 1M and 10M?

tevador:
Some notes about hardware AES:

RISC-V actually has 2 different crypto extensions:

1. Scalar crypto, which uses the integer registers. The extension was [ratified in January 2022](https://github.com/riscv/riscv-crypto/releases/tag/v1.0.0-scalar). The first chips with the support of this extension might appear next year.
2. Vector crypto, which depends on the V extension and uses the vector registers. It was [ratified yesterday](https://github.com/riscv/riscv-crypto/releases/tag/v1.0.0). It will probably take a couple years before we see the first chips that support it.

So it's possible that we'll have to support at least 2 different extensions in the future with the scalar one likely coming first. Actually, the scalar crypto AES instructions are split into two separate extensions Zkne (encryption only) and Zknd (decryption only). Hopefully, hardware designers will be sane and include both extensions.

In order to limit the scope of this PR, I did not include hardware AES for now. It would not be very useful anyways since no chips you can buy today support it.

SChernykh:
> I'm running 1M now (should take a couple hours) and then I'll try 10M. Then we'll have to repeat the runs with the native build. Do you have the correct hashes for 1M and 10M?

1M `b4b51690546a44a459ad3e043369e1237f61c5ec27046d2f5249cf0b3f57e00c`
10M `c32b1910abef45bc659ffd8d5aeeb0384f7122e623f2fafe568a393a9c3be60b`

tevador:
```
> ./randomx-benchmark --auto --mine --largePages --threads 4 --nonces 1000000
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing (4 threads) ...
Memory initialized in 44.2732 s
Initializing 4 virtual machine(s) ...
Running benchmark (1000000 nonces) ...
Calculated result: b4b51690546a44a459ad3e043369e1237f61c5ec27046d2f5249cf0b3f57e00c
Performance: 72.6978 hashes per second
```

tevador:
10M also matches with the default build

```
> ./randomx-benchmark --auto --mine --largePages --threads 4 --nonces 10000000
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing (4 threads) ...
Memory initialized in 44.2651 s
Initializing 4 virtual machine(s) ...
Running benchmark (10000000 nonces) ...
Calculated result: c32b1910abef45bc659ffd8d5aeeb0384f7122e623f2fafe568a393a9c3be60b
Performance: 72.7311 hashes per second
```

SChernykh (approved):

felixonmars:
Tried on more boards, all built with default rv64gc, with comparison to x86_64:

SChernykh:
@felixonmars How many threads did you run on SG2042? The optimal number of threads is 32 there.

hyc:
Was going to say the same. If that run was with 64 threads, need to try again with 32.

felixonmars:
> @felixonmars How many threads did you run on SG2042? The optimal number of threads is 32 there.

I did run with 64. The results vary quite much on each run and even down to only 2 hps sometimes. I'll retry with 32 later (it's fully loaded now).

tevador:
@felixonmars I can see from your JH7110 hashrates that you are probably not using hugepages. Try to use `--largePages` when running the benchmark. It can boost the performance by 30%.

I ran the 10M hashes with the native build and the result also matches.

```
> ./randomx-benchmark --auto --mine --largePages --threads 4 --nonces 10000000
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing (4 threads) ...
Memory initialized in 41.2095 s
Initializing 4 virtual machine(s) ...
Running benchmark (10000000 nonces) ...
Calculated result: c32b1910abef45bc659ffd8d5aeeb0384f7122e623f2fafe568a393a9c3be60b
Performance: 74.9915 hashes per second
```

I will wait for someone to independently verify my hashes before merging this PR.

felixonmars:
> I can see from your JH7110 hashrates that you are probably not using hugepages. Try to use `--largePages` when running the benchmark. It can boost the performance by 30%.

Indeed. I tried with `--largePages` but the results are much lower for SG2042. And my TH1520 kernel doesn't have support for hugepages. So I disabled it for all my benchmarks :(

hyc:
Got some results on Lichee Pi 4a. Numbers were quite slow on the shipped firmware, dated July 2023. Updated to September 20 2023 image and my results look more reasonable. Speed with 1 thread and no largePages was 34.89H/s, same as @felixonmars got. With largepages
 1 thread = 43.45H/s
 2 thread = 78.71H/s
 4 thread = 132.54H/s

felixonmars:
> I will wait for someone to independently verify my hashes before merging this PR.

```
$ ./randomx-benchmark --auto --mine --largePages --threads 32 --nonces 10000000
RandomX benchmark v1.1.12
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - software AES mode
 - large pages mode
 - batch mode
Initializing (64 threads) ...
Memory initialized in 7.06713 s
Initializing 32 virtual machine(s) ...
Running benchmark (10000000 nonces) ...
Calculated result: c32b1910abef45bc659ffd8d5aeeb0384f7122e623f2fafe568a393a9c3be60b
Performance: 424.954 hashes per second
```

Hashes match here. With large pages on and threads set to 32, the performance is much lower though.

hyc (approved):

hyc:
> > I will wait for someone to independently verify my hashes before merging this PR.
> 
> ```
> $ ./randomx-benchmark --auto --mine --largePages --threads 32 --nonces 10000000
> RandomX benchmark v1.1.12
>  - Argon2 implementation: reference
>  - full memory mode (2080 MiB)
>  - JIT compiled mode
>  - software AES mode
>  - large pages mode
>  - batch mode
> Initializing (64 threads) ...
> Memory initialized in 7.06713 s
> Initializing 32 virtual machine(s) ...
> Running benchmark (10000000 nonces) ...
> Calculated result: c32b1910abef45bc659ffd8d5aeeb0384f7122e623f2fafe568a393a9c3be60b
> Performance: 424.954 hashes per second
> ```
> 
> Hashes match here. With large pages on and threads set to 32, the performance is much lower though.

Since the C920 caches are arranged in clusters per 4 cores, you'd need to be able to pin thread and memory allocations to particular cores to obtain optimal memory layout. Along the lines of what the numactl command does.

Issue #277, opened 2023-10-09
Compilation failed in a project using RandomX


LithiumTins:
I encountered a problem when compiling a project that uses RandomX. One of the files, which includes 'intrin_portable.h', fails to compile due to '-Werror=cast-qual'. Here is the relevant code.
https://github.com/tevador/RandomX/blob/1c603a23fa0004615b4f2d42ad95f2693f16093b/src/intrin_portable.h#L709-L714
It seems that casting 'addr' to the 'uint8_t*' cause this error. However, the function load32() accepts a parameter of type 'const uint8_t *'. 
https://github.com/tevador/RandomX/blob/1c603a23fa0004615b4f2d42ad95f2693f16093b/src/blake2/endian.h#L29
I just wonder whether it would be fine to cast 'addr' to 'const uint8_t *' instead of 'uint8_t *', for example,
```cpp
x.lo = (double)unsigned32ToSigned2sCompl(load32((const uint8_t*)addr + 0)); 
```

SChernykh:
Yes, it's totally fine. I even fixed it in my P2Pool branch: https://github.com/SChernykh/RandomX/commit/7167cf1ce6e5e9b240967e0d96148966a24bbdb0
You can make a pull request here - I didn't make it because it didn't affect anyone else.

LithiumTins:
> Yes, it's totally fine. I even fixed it in my P2Pool branch: [SChernykh@7167cf1](https://github.com/SChernykh/RandomX/commit/7167cf1ce6e5e9b240967e0d96148966a24bbdb0) You can make a pull request here - I didn't make it because it didn't affect anyone else.

Actually I just met the problem when I tried to build the archlinux package of your P2Pool from on RISCV64 :)
It seems you've fixed this on aarch64 part，would you mind fix this on other parts of intrin_portable.h?

SChernykh:
Ok, I fixed it in P2Pool

PR #278, opened 2023-10-09
Preserve const qualifier when casting

tevador:
#277


SChernykh:
I found several more places where it can be const: https://github.com/SChernykh/RandomX/commit/d5035cd9c50a5d34ae87ec3507314dc806343548

SChernykh (approved):

PR #279, opened 2023-10-09
Avoid redundant CI runs

tevador:
Currently, CI jobs run twice for every push to a branch that has an open pull request.

This should limit CI runs only to pushes to master or PR ~~opened against master~~.


SChernykh:
```
on:
  push:
    branches: [ master ]
  pull_request:
```
Should be better, it will allow CI on pull requests to other branches.

SChernykh (approved):

PR #281, opened 2023-10-19
ARM64 JIT: don't use `x18` register

SChernykh:
From https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms
> The platforms reserve register x18. Don’t use this register.

This PR fixes invalid hashes when running on Apple silicon with the latest macOS SDK.


selsta:
`./randomx-tests` and `./randomx-benchmark --mine --init 10 --threads 10 --jit --nonces 1000000` both pass now with the latest macOS SDK.

tevador (approved):

hyc (approved):
Still no problems on M1 Macbook pro running MacOS 12.6.5, build 21G531, clang 13.0

PR #282, opened 2023-10-20
Fixed UB in ARM64 JIT compiler

SChernykh:
Fixed unaligned memory writes

### DIFF
--- a/src/jit_compiler_a64.cpp
+++ b/src/jit_compiler_a64.cpp
@@ -709,7 +709,8 @@ void JitCompilerA64::h_IMUL_RCP(Instruction& instr, uint32_t& codePos)
 	const uint32_t literal_id = (ImulRcpLiteralsEnd - literalPos) / sizeof(uint64_t);
 
 	literalPos -= sizeof(uint64_t);
-	*(uint64_t*)(code + literalPos) = (q << shift) + ((r << shift) / divisor);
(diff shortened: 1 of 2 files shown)


src/jit_compiler_a64.cpp
```diff
@@ -709,7 +709,8 @@ void JitCompilerA64::h_IMUL_RCP(Instruction& instr, uint32_t& codePos)
 	const uint32_t literal_id = (ImulRcpLiteralsEnd - literalPos) / sizeof(uint64_t);
 
 	literalPos -= sizeof(uint64_t);
-	*(uint64_t*)(code + literalPos) = (q << shift) + ((r << shift) / divisor);
+	const uint64_t literal = (q << shift) + ((r << shift) / divisor);
```
tevador, line 712: IMO this should call `randomx_reciprocal` for compatibility reasons so we don't have the same code scattered in multiple places.

SChernykh: `randomx_reciprocal` is a slow function. Even x86 JIT calls `randomx_reciprocal_fast`. This is the fast version for ARM64.

SChernykh: I agree it would be better to move this code to the reference `randomx_reciprocal`, but it's not the scope of this PR.

tevador (approved):

Issue #153, opened 2019-11-12
Failed to link librandomx.a in monero mobile wallet for ios 


naughtyfox:
Hi! I'm sorry for duplicating the issue from monero repo https://github.com/monero-project/monero/issues/6122, but I think this is more relevant place for this.

We are trying to build ios exawallet with new release libraries. For x86-64 simulator it builds fine, but the linking problem occurs for arm64:
```
Undefined symbols for architecture arm64:
  "_randomx_program_aarch64_main_loop", referenced from:
      __GLOBAL__sub_I_jit_compiler_a64.cpp in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_program_aarch64_vm_instructions", referenced from:
      __GLOBAL__sub_I_jit_compiler_a64.cpp in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_program_aarch64_imul_rcp_literals_end", referenced from:
      __GLOBAL__sub_I_jit_compiler_a64.cpp in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_init_dataset_aarch64_end", referenced from:
      __GLOBAL__sub_I_jit_compiler_a64.cpp in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_init_dataset_aarch64", referenced from:
      randomx::JitCompilerA64::getDatasetInitFunc() in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_program_aarch64_vm_instructions_end_light", referenced from:
      randomx::JitCompilerA64::generateProgramLight(randomx::Program&, randomx::ProgramConfiguration&, unsigned int) in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_program_aarch64_light_cacheline_align_mask", referenced from:
      randomx::JitCompilerA64::generateProgramLight(randomx::Program&, randomx::ProgramConfiguration&, unsigned int) in librandomx.a(jit_compiler_a64.cpp.o)
  "_randomx_calc_dataset_item_aarch64_store_result", referenced from:
      void randomx::JitCompilerA64::generateSuperscalarHash<8ul>(randomx::SuperscalarProgram (&) [8ul], std::__1::vector<unsigned long long, std::__1::allocator<unsigned long long> >&) in librandomx.a(jit_compiler_a64.cpp.o)
      __GLOBAL__sub_I_jit_compiler_a64.cpp in librandomx.a(jit_compiler_a64.cpp.o)
[...]

PR #294, opened 2024-04-02
fix missing ___clear_cache when targetting iOS

MrCyjaneK:
Without the patch I get the following issue

```
Undefined symbols for architecture arm64:
  "___clear_cache", referenced from:
      void randomx::JitCompilerA64::generateSuperscalarHash<8ul>(randomx::SuperscalarProgram (&) [8ul], std::__1::vector<unsigned long long, std::__1::allocator<unsigned long long>>&) in librandomx.a[27](jit_compiler_a64.cpp.o)
      randomx::JitCompilerA64::JitCompilerA64() in librandomx.a[27](jit_compiler_a64.cpp.o)
      randomx::JitCompilerA64::generateProgram(randomx::Program&, randomx::ProgramConfiguration&) in librandomx.a[27](jit_compiler_a64.cpp.o)
      randomx::JitCompilerA64::generateProgramLight(randomx::Program&, randomx::ProgramConfiguration&, unsigned int) in librandomx.a[27](jit_compiler_a64.cpp.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I've found the very same fix used in https://github.com/xmrig/xmrig/commit/10d292092a45941da4f79a0cd029705a34bd713b

I didn't check the fix on iOS yet, but the library linked properly, so I guess it is good.


SChernykh:
This is not a fix, you just removed (using #ifdef) the code that breaks the build. You still need to clear the code cache on arm64 CPUs. If you read the discussion in https://github.com/xmrig/xmrig/issues/1246 and check the code in https://github.com/xmrig/xmrig/blob/master/src/crypto/common/VirtualMemory_unix.cpp#L206 you'll see there are more things needed to be done here to fix it.

I'm not sure this fix is even needed because all recent clang versions have `__GNUC__` and `__builtin___clear_cache` defined and automatially translate it to the `sys_icache_invalidate` call.

MrCyjaneK:
> This is not a fix, you just removed (using #ifdef) the code that breaks the build. You still need to clear the code cache on arm64 CPUs. If you read the discussion in https://github.com/xmrig/xmrig/issues/1246 and check the code in https://github.com/xmrig/xmrig/blob/master/src/crypto/common/VirtualMemory_unix.cpp#L206 you'll see there are more things needed to be done here to fix it.

Sorry! I've missed that part of the discussion, is fix like this more suitable or should I leave `__GNUC__` check in place?

```
#ifdef HAVE_BUILTIN_CLEAR_CACHE
__builtin___clear_cache(reinterpret_cast<char*>(code), reinterpret_cast<char*>(code + CodeSize));
#else
sys_icache_invalidate(code, code + CodeSize);
#endif
```

> I'm not sure this fix is even needed because all recent clang versions have __GNUC__ and __builtin___clear_cache defined and automatially translate it to the sys_icache_invalidate call.

I'm currently building with the toolchain that ships with latest Xcode on macos, so at least on my machine it didn't work.

SChernykh:
The C++ code has `__builtin___clear_cache` but the linker complains about `___clear_cache` so the clang compiler does know about this function and replaces it with the call to `__clear_cache`, the problem is that the code is not linked against the correct library (libgcc probably). IMHO even the XMRig's solution is not a proper one, it just works around the problem which is somewhere in the CMake build step. Can you try https://github.com/iains/gcc-darwin-arm64/issues/33#issuecomment-751756730 and other suggestions there?

MrCyjaneK:
I've seen these solutions (and tried some), and they target macos instead of ios, and according to (a rather old) issue https://gitlab.haskell.org/ghc/ghc/-/issues/8561 ___clear_cache is a thing on MacOS but not iOS, people there also mention that JIT is not really a possible option on non-jailbroken iOS device, which also makes RandomX unusable, at least for the vast majority of iOS users.

luajit case mentioned in that issue also seems to treat iOS as a special thing when it comes to clearning cache https://repo.or.cz/luajit-2.0.git/blob/HEAD:/src/lj_mcode.c#l51

```
#if LJ_TARGET_X86ORX64
  UNUSED(start); UNUSED(end);
#elif LJ_TARGET_WINDOWS
  FlushInstructionCache(GetCurrentProcess(), start, (char *)end-(char *)start);
#elif LJ_TARGET_IOS
  sys_icache_invalidate(start, (char *)end-(char *)start);
#elif LJ_TARGET_PPC
  lj_vm_cachesync(start, end);
#elif defined(__GNUC__) || defined(__clang__)
  __clear_cache(start, end);
#else
#error "Missing builtin to flush instruction cache"
#endif
```

src/jit_compiler_a64.cpp
```diff
@@ -98,8 +101,10 @@ JitCompilerA64::JitCompilerA64()
 	memset(reg_changed_offset, 0, sizeof(reg_changed_offset));
 	memcpy(code, (void*) randomx_program_aarch64, CodeSize);
 
-#ifdef __GNUC__
-	__builtin___clear_cache(reinterpret_cast<char*>(code), reinterpret_cast<char*>(code + CodeSize));
+#ifdef HAVE_BUILTIN_CLEAR_CACHE
+ 	__builtin___clear_cache(reinterpret_cast<char*>(code), reinterpret_cast<char*>(code + CodeSize));
+#else
+ 	sys_icache_invalidate(code, (size_t)(code + CodeSize));
```
SChernykh, line 107: The definition of the function is `sys_icache_invalidate(void *start, size_t len);` so it must be
```
sys_icache_invalidate(code, CodeSize);
```
here and in other places.

MrCyjaneK: Thanks, done.

MrCyjaneK:
> Do not merge, aarch64-linux-android was broken in this port, I'll take look at it in a bit.

Resolved. Testing on iOS is still pending - but it should work without issues, it doesn't break windows, linux and macos builds.

tevador:
Related issue: https://github.com/tevador/RandomX/issues/153

PR #299, opened 2024-08-12
More faster init cache creation

ilnarildarovuch:
Refactored randomx_init_dataset function in src/randomx.cpp


ilnarildarovuch:
Night, dark night

PR #300, opened 2024-08-12
More faster init cache creation

ilnarildarovuch:
Refactored randomx_init_dataset function in src/randomx.cpp -> running faster, than normal about .8 seconds.
Tested on Windows 11.

### DIFF
--- a/src/randomx.cpp
+++ b/src/randomx.cpp
@@ -127,14 +127,15 @@ extern "C" {
 		return cache;
 	}
 
-	void randomx_init_cache(randomx_cache *cache, const void *key, size_t keySize) {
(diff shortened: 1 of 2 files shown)


ilnarildarovuch:
Retested 30 times, with some apps open and closed, result -> 0.2

SChernykh:
`src/vm_compiled.cpp` is only whitespace changes, it shouldn't be here. Keep the diff is small as possible - don't change whitespace or move `*` around, because it breaks commit history and `git blame` functionality.

`randomx_init_cache` takes at most 0.0001 seconds itself, the actual initialization is done inside `cache->initialize`. Removing a single copy of a short string doesn't make any real difference.

It is very unlikely to be merged for these reasons.

ilnarildarovuch:
Ok

PR #317, opened 2026-01-26
RandomX v2

SChernykh:
RandomX v2 includes the following changes:

- Program size increased from 256 to 384
- CFROUND is tweaked to change the rounding 16 times less often
- Group F and group E register mix is done using 16 AES operations (the extra 262144 AES ops come from this change)
- Dataset prefetch is 2 iterations ahead instead of 1

RandomX v1 hash: 4194304 VM instructions + 262144 AES = 4456448 total ops
RandomX v2 hash: 6291456 VM instructions + 524288 AES = 6815744 total ops (+52.9%)

See [design_v2.md](https://github.com/SChernykh/RandomX/blob/v2/doc/design_v2.md) and search for "RandomX v2" in [specs.md](https://github.com/SChernykh/RandomX/blob/v2/doc/specs.md) for more details on the tweaks.


SChernykh:
@tevador @hyc Please review, it's been almost 3 weeks.

.github/workflows/c-cpp.yml
```diff
@@ -62,6 +62,7 @@ jobs:
       matrix:
         config:
           - {os: ubuntu-24.04, c: gcc-14, cpp: g++-14}
+          - {os: ubuntu-24.04-arm, c: gcc-14, cpp: g++-14}
```
hyc, line 65: Seems like changes to this file are unrelated to v2. Can this be a separate PR? This is already a large PR with 25 commits.

doc/design_v2.md
```diff
@@ -0,0 +1,111 @@
...
+- AES tweak also improves data entropy (makes it more random) before it's written to the scratchpad.
+
+## 3. Program size increase from 256 to 384
+
+CPUs got much faster since the original RandomX was released. Back in 2019, Ryzen 9 3950X was the fastest desktop CPU for RandomX, and at the time of writing (January 2026) it's Ryzen 9 9950X. In most CPU benchmarks, 9950X is more than 1.5x faster on average - thanks to clock speed increase from < 4 to > 5 GHz, and to IPC improvements.
+
+But in RandomX it's only 20-25% faster, because it's bottlenecked by the RAM latency. While CPU cores got faster over the years, RAM latency stayed basically the same - a tuned DDR4 memory from 2019, and a tuned DDR5 memory from 2026 will both have the same access latency of around 50-55 ns.
+
+This disbalance is the main reason of the program size increase - Zen5 and newer CPUs need more work to keep themselves busy while they're waiting for data from memory.
```
hyc: s/disbalance/imbalance/

src/asm/program_epilogue_linux.inc
```diff
@@ -1,10 +1,12 @@
 	;# restore callee-saved registers - System V AMD64 ABI
-	pop r15
```
hyc, line 2: these prologue/epilogue changes were required to reserve more stack space?

src/intrin_portable.h
```diff
@@ -276,6 +276,14 @@ FORCE_INLINE rx_vec_f128 rx_set1_vec_f128(uint64_t x) {
 	return (rx_vec_f128)vec_splat2sd(x);
 }
 
+FORCE_INLINE rx_vec_f128 rx_cast_vec_i2f(rx_vec_i128 a) {
```
hyc, line 279: these changes look like a bugfix, not enabling CI. ?

.github/workflows/c-cpp.yml
SChernykh: I added this config because it runs on native ARM64 hardware - it's important to test everything not just in QEMU.

src/asm/program_epilogue_linux.inc
SChernykh: Yes, more stack is required for software AES code in the main loop

src/intrin_portable.h
SChernykh: This is a bugfix (one of) that was required to make CI all green.

hyc (dismissed):

doc/specs.md
```diff
@@ -447,12 +450,14 @@ The loop described below is repeated until the value of the `ic` register reache
 2. `spAddr0` is used to perform a 64-byte aligned read from Scratchpad level 3 (using mask from Table 4.2.1). The 64 bytes are XORed with all integer registers in order `r0`-`r7`.
 3. `spAddr1` is used to perform a 64-byte aligned read from Scratchpad level 3 (using mask from Table 4.2.1). Each floating point register `f0`-`f3` and `e0`-`e3` is initialized using an 8-byte value according to the conversion rules from chapters 4.3.1 and 4.3.2.
 4. The 256 instructions stored in the Program Buffer are executed.
-5. The `mx` register is XORed with the low 32 bits of registers `readReg2` and `readReg3` (see Table 4.5.3).
-6. A 64-byte Dataset item at address `datasetOffset + mx % RANDOMX_DATASET_BASE_SIZE` is prefetched from the Dataset (it will be used during the next iteration).
-7. A 64-byte Dataset item at address `datasetOffset + ma % RANDOMX_DATASET_BASE_SIZE` is loaded from the Dataset. The 64 bytes are XORed with all integer registers in order `r0`-`r7`.
+5. The value of `ma` is saved as a temporary `t`. Then the `mp` register is XORed with the low 32 bits of registers `readReg2` and `readReg3` (see Table 4.5.3).
```
tevador: Nitpick: Register `t` should be explicitly defined as an additional VM register on line 322. Perhaps it could be renamed to something like `mt`.

src/jit_compiler_a64.cpp
```diff
@@ -203,19 +241,56 @@ void JitCompilerA64::generateProgramLight(Program& program, ProgramConfiguration
...
+		}
+		else {
+			// Jump to RandomX v2 FE mix soft AES code by writing "b randomx_program_aarch64_v2_FE_mix_soft_aes" instruction
+			uint32_t offset = (uint8_t*)randomx_program_aarch64_v2_FE_mix_soft_aes - (uint8_t*)randomx_program_aarch64_v2_FE_mix;
+			emit32(ARMV8A::B | (offset / 4), code, codePos);
+
+			offset = (uint8_t*)randomx_program_aarch64_aes_lut_pointers - (uint8_t*)randomx_program_aarch64;
+
+			*(uint64_t*)(code + offset + 0) = (uint64_t) &randomx_aes_lut_enc[0][0];
+			*(uint64_t*)(code + offset + 8) = (uint64_t) &randomx_aes_lut_dec[0][0];
```
tevador: I think this is technically UB due to the strict aliasing rule.

src/jit_compiler_rv64_vector.cpp
```diff
@@ -330,6 +334,24 @@ void* generateProgramVectorRV64(uint8_t* buf, Program& prog, ProgramConfiguratio
...
+		params[7] = (uint64_t) randomx_aes_lut_enc_index;
+		params[8] = (uint64_t) randomx_aes_lut_dec_index;
+
+		uint32_t* p1 = (uint32_t*)(buf + DIST(randomx_riscv64_vector_code_begin, randomx_riscv64_vector_program_v2_soft_aes_init));
+
+		// Restore vsetivli zero, 4, e32, m1, ta, ma
+		*p1 = 0xCD027057;
+	}
+	else {
+		uint32_t* p1 = (uint32_t*)(buf + DIST(randomx_riscv64_vector_code_begin, randomx_riscv64_vector_program_v2_soft_aes_init));
```
tevador, line 349: Strict aliasing.

src/jit_compiler_rv64_vector.cpp
```diff
@@ -887,8 +885,26 @@ void* generateProgramVectorRV64(uint8_t* buf, Program& prog, ProgramConfiguratio
 		e = buf + DIST(randomx_riscv64_vector_code_begin, randomx_riscv64_vector_program_main_loop_instructions_end);
 	}
 
-	const uint32_t k = e - p;
-	emit32(0x6F | ((k & 0x7FE) << 20) | ((k & 0x800) << 9) | (k & 0xFF000));
+	emit32(JUMP(e - p));
+
+	if (flags & RANDOMX_FLAG_V2) {
+		uint32_t* p1 = (uint32_t*)(buf + DIST(randomx_riscv64_vector_code_begin, randomx_riscv64_vector_program_main_loop_fe_mix));
```
tevador, line 891: Strict aliasing.

src/jit_compiler_a64.cpp
SChernykh: In this specific case it's not related to strict aliasing, it's a different thing - a "pointer-to-integer" conversion, which is not UB but implementation defined (meaning that the implementation defines the exact bit contents of `uint64_t`). Chances are close to 0 that it will go wrong because this file is compiled only on 64-bit systems and uses 64-bit types, they are large enough to store pointers.

SChernykh: Or do you mean the left side, where `uint8_t*` is accessed as `uint64_t*`? Yes, it's probably unsafe due to possible alignment issues. I need to use `memcpy` instead.

tevador: Yes, I mean the cast `uint8_t*` -> `uint64_t*`.

tevador (approved):

src/jit_compiler_a64.cpp
```diff
@@ -185,8 +185,11 @@ void JitCompilerA64::generateProgram(Program& program, ProgramConfiguration& con
 
 			offset = (uint8_t*)randomx_program_aarch64_aes_lut_pointers - (uint8_t*)randomx_program_aarch64;
 
-			*(uint64_t*)(code + offset + 0) = (uint64_t) &randomx_aes_lut_enc[0][0];
-			*(uint64_t*)(code + offset + 8) = (uint64_t) &randomx_aes_lut_dec[0][0];
+			const void* lut_enc = &randomx_aes_lut_enc[0][0];
+			const void* lut_dec = &randomx_aes_lut_dec[0][0];
+
+			memcpy(code + offset + 0, &lut_enc, sizeof(lut_enc));
+			memcpy(code + offset + 8, &lut_dec, sizeof(lut_dec));
```
hyc, line 192: why use '8' and 'sizeof'? if you're using sizeof, then it doesn't make sense to use a magic number 8. If you know that these types are 64bit, then there's no need to use sizeof. Since this is platform-specific arm64, you know the type sizes...

SChernykh: Because it's defined as 2x8 byte field in the asm file, hence +8. `memcpy(..., &x, sizeof(x))` is just a standard idiom that makes code analyzers and compilers (at high warning level) happy. I know that it's all the same in this platform-specific code.

hyc: Okeydoke

hyc (approved):

PR #320, opened 2026-04-04
Add JIT code generator for PPC64

### LINKED ISSUE #132: POWER support
Notably absent from the list of optimized architectures is any of the open ISAs.  For integrity reasons it would make sense to support at least one of those ISAs in a proper mining role.

I propose the POWER ISA be used as reasonably powerful CPUs are commercially available that implement it.

Is there any interest in doing this and what would be needed to make it happen?

cyrozap:
Adds a JIT backend for POWER8 and later Power ISA CPUs. Assembly instructions were restricted to those available in Power ISA v2.06 in order to facilitate adding support for POWER7, but currently only RandomX V1 is supported on those chips due to its lack of AES instructions.

Support has been added for both little-endian and big-endian CPUs, but only little-endian has been tested.

Fixes #132


cyrozap:
Benchmarks on a Raptor Computing Systems Talos II with dual POWER9 CPUs:

`./randomx-benchmark --auto --verify`:

```
RandomX benchmark v2.0
 - Argon2 implementation: reference
 - light memory mode (256 MiB)
 - JIT compiled mode
 - hardware AES mode
 - small pages mode
 - batch mode
Initializing ...
Memory initialized in 1.01555 s
Calculated result: 10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Reference result:  10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Performance: 25.3256 ms per hash
```

`./randomx-benchmark --auto --verify --v2`:

```
RandomX benchmark v2.0
 - Argon2 implementation: reference
 - light memory mode (256 MiB)
 - JIT compiled mode
 - hardware AES mode
 - small pages mode
 - batch mode
Initializing ...
Memory initialized in 1.0161 s
Calculated result: b85d79e080b10b6ad28c2e6c993601a1361917dba979e03a0a8f7248aaf4ba52
Reference result:  b85d79e080b10b6ad28c2e6c993601a1361917dba979e03a0a8f7248aaf4ba52
Performance: 26.8799 ms per hash
```

`./randomx-benchmark --auto --mine`:

```
RandomX benchmark v2.0
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - hardware AES mode
 - small pages mode
 - batch mode
Initializing (144 threads) ...
Memory initialized in 2.69269 s
Calculated result: 10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Reference result:  10b649a3f15c7c7f88277812f2e74b337a0f20ce909af09199cccb960771cfa1
Performance: 237.531 hashes per second
```

`./randomx-benchmark --auto --mine --v2`:

```
RandomX benchmark v2.0
 - Argon2 implementation: reference
 - full memory mode (2080 MiB)
 - JIT compiled mode
 - hardware AES mode
 - small pages mode
 - batch mode
Initializing (144 threads) ...
Memory initialized in 2.71528 s
Calculated result: b85d79e080b10b6ad28c2e6c993601a1361917dba979e03a0a8f7248aaf4ba52
Reference result:  b85d79e080b10b6ad28c2e6c993601a1361917dba979e03a0a8f7248aaf4ba52
Performance: 177.077 hashes per second
```

SChernykh:
> due to its lack of AES instructions.

RandomX v1 also uses AES in the scratchpad hash/fill step, so you can use the existing soft AES code for RandomX v2 loop. It shouldn't be that hard compared to the full JIT implementation that you've done already.

tevador:
FYI, the ppc64le build is failing: https://github.com/tevador/RandomX/actions/runs/24008647458/job/70025217444

cyrozap:
> FYI, the ppc64le build is failing: https://github.com/tevador/RandomX/actions/runs/24008647458/job/70025217444

From the CI log:

```
/home/runner/work/RandomX/RandomX/src/cpu.cpp:52:18: fatal error: asm/cputable.h: No such file or directory
   52 |         #include <asm/cputable.h>
      |                  ^~~~~~~~~~~~~~~~
```

`asm/cputable.h` is a Linux kernel header, so I think you would need to run `apk add linux-headers` in the ppc64le VM setup script to resolve that error.

We could also avoid the dependency entirely by `#define`-ing the value ourselves. ~This constant is needed in both `src/cpu.cpp` and `src/jit_compiler_ppc64.cpp`, so if you want to do this we'd have to make that change in both places (or make our own header that gets included into both files).~ **Edit**: I just realized I didn't actually need to call `getauxval` a second time, so I pushed a change to fix that. Now the constant is only needed in `src/cpu.cpp` and in the JIT backend we just query the `cpu` object for feature support when we need it.

Which option would you prefer?

FWIW, in the future I plan to use more of those definitions (full list is [here](https://www.kernel.org/doc/html/next/arch/powerpc/elf_hwcaps.html)) to detect the system's ISA version in order to patch in more-optimized code for the newer architectures, so my personal preference is to just use the Linux kernel header so there's no possibility of copy/paste issues. That said, I'll understand completely if you want to avoid a dependency on kernel headers just for a handful of constant definitions (which AIUI should never change between kernel versions).

cyrozap:
Just a quick update on this: Big-endian PPC64 works now, but after doing more thorough testing on little-endian, it seems there's a very intermittent bug where the scratchpad pointer base address is getting added to `spAddr0` / `spAddr1` twice, resulting in out-of-bounds writes with associated segfaults and data corruption. I'll let you know once I've found the root cause of the issue and fixed it.

cyrozap:
I fixed the segfaults and memory corruption bugs. Turns out the caches weren't being invalidated after writing the code, causing stale instructions to get executed from the cache and leading to intermittent, unpredictable results. But now the caches are being properly invalidated and the code is running reliably.

cyrozap:
@tevador I hadn't heard back from you regarding what you wanted to do about the missing kernel header issue, so in 9a77acf I just went ahead and removed the need for that dependency entirely. So if you want to run CI on this PR, it should be working now.

Separately, I got software AES working, so RandomX V2 should be able to run on POWER7 now. I've tested the implementation and it works both on my little-endian POWER9 system and in a Qemu-emulated big-endian POWER7 system.

At this point, I consider this implementation basically "complete" (pending review, of course). It works on (emulated) POWER7 through (bare-metal) POWER9, on both big-endian and little-endian systems. There's things that could be optimized in the software AES implementation, but I've decided to hold off on doing any of that until I can get access to a real POWER7 system to validate that any "improvements" I make actually increase performance.

Anyways, if there's anything I need to change or fix, please let me know!

cyrozap:
I pushed a fix for the missing sysconf definitions (and also a 2% perf increase)--please run CI again at your earliest convenience.

SChernykh:
@cyrozap Is it complete, or are you planning to add more commits? Can you paste the randomx-benchmark results for the latest code?

cyrozap:
@SChernykh This is complete, at least in the near term--I'm pretty sure the generated POWER9 code is now nearly as optimized as it can be, and I won't be able to make any improvements specific to POWER7 and POWER8 for probably another six months or more as I don't yet have hardware for that set up.

Unfortunately, I can't provide randomx-benchmark results at the moment as I don't currently have access to my POWER9 system, but I'll likely be able to in about two months. That said, the last benchmarks I did with xmrig I was able to get about 309 H/s per thread on average for RandomX V1 and about two-thirds of that value for V2.

SChernykh:
@cyrozap please rebase to the current master branch, I see that CI misses some tests that were added recently.

src/jit_compiler_ppc64.cpp
```diff
@@ -0,0 +1,1622 @@
...
+	static void emitMovImm64(CompilerState& state, int reg, uint64_t imm) {
+		if (imm == (uint64_t)(int64_t)(int32_t)imm) {
+			// Values that can be represented by loading a 32-bit signed immediate
+			emitMovImm32(state, reg, (uint32_t)imm);
+		} else {
+			uint64_t lowestBit = imm & -(int64_t)imm;
+			uint64_t added = imm + lowestBit;
+			if (imm != 0 && imm != ~0ULL && (added & (added - 1)) == 0) {
+				// Values that are a contiguous sequence of 1s
+				uint32_t mb = added == 0 ? 0 : __builtin_clzll(added);
```
SChernykh, line 746: Off-by-one error here: imm = 0xFFFFFFFF (2^32 - 1) will generate code that loads 0x1FFFFFFFF (2^33 + 2^32 - 1).
Fix:
```
uint32_t mb = (added == 0) ? 0 : (__builtin_clzll(added) + 1);
```

src/jit_compiler_ppc64_static.S
```diff
@@ -0,0 +1,1389 @@
...
+// v27 / vs59 (non-volatile) -> subbytes_mul1_hi / invsubbytes_mul9_hi
+// v28 / vs60 (non-volatile) -> subbytes_mul2_lo / invsubbytes_mulD_lo
+// v29 / vs61 (non-volatile) -> subbytes_mul2_lo / invsubbytes_mulD_hi
+// v30 / vs62 (non-volatile) -> invsubbytes_mulB_lo
+// v31 / vs63 (non-volatile) -> invsubbytes_mulB_hi
+
+randomx_ppc64_vm_spad_store_mix_v2_soft_aes:
+	// Mix F and E registers (f0-f3 are v0-v3, e0-e3 are v4-v7)
+
+	// Save v8-v11 and v15-v31 to the stack
```
SChernykh, line 1102: This doesn't allocate the actual stack frame, so you're saving data to the red zone. If a signal arrives, the signal handler can overwrite this. You should allocate the stack frame by doing `stdu r1, -352(r1)` … `addi r1, r1, 352`

SChernykh:
@cyrozap Looks like your JIT code has the same bug with ISUB_R that was found for ARM/RISC-V implementations earlier.

PR #325, opened 2026-05-08
ISUB_R fix for ARM/RISC-V JIT

SChernykh:
When `src = dst` and `imm = 0x80000000` in ISUB_R, the JIT generated the wrong code because negating 0x80000000 (-2^31) resulted in -2^31 again, not +2^31.

Also added the test to `tests.cpp` to check this case (can be verified by disabling the fix in jit_compiler_a64/jit_compiler_rv64).


tevador:
Tested and reproduced on a raspberry pi.

Just one small nitpick: please change 40 to 44 in tests.cpp on line 62 (alignment).

SChernykh:
@tevador fixed

tevador (approved):

PR #331, opened 2026-05-24
cmake: set minimum to 3.10

tobtoht:
```
CMake Deprecation Warning at CMakeLists.txt:29 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.
```

This will eventually break downstream builds, so update it before compatibility is removed.


SChernykh (approved):

SChernykh:
@tobtoht Can you make this PR against v1.x branch?

